PR #2, opened 2019-10-17
fix for time units attribute when IAU used with netcdf format

jswhit2:
Port of https://vlab.ncep.noaa.gov/redmine/issues/69735

When `iau_offset`> 0, the forecast hour is reset but the initial time is not reset when netcdf output is enabled. This update modifies the `units` attribute of the time variable in the output netcdf files when `iau_offset` > 0. This update is necessary for IAU to work with netcdf io enabled.


io/module_write_netcdf.F90
```diff
@@ -555,8 +557,73 @@ subroutine add_dim(ncid, dim_name, dimid, grid, rc)
...
+! rinc: days,hours,minutes,seconds,milliseconds
+       rinc = 0; rinc(2) = iau_offset
+       call w3movdat(rinc,idat,jdat)
+! update idate using iau_offset
+       idate(1:3)=jdat(1:3)
+       idate(4:6)=jdat(5:7)
+       time_units  = get_time_units_from_idate(idate)
+! rewrite time units attribute.
+       ncerr = nf90_put_att(ncid, dim_varid, 'units', trim(time_units)); NC_ERR_STOP(ncerr)
+    endif
```
DusanJovic-NOAA: Can this manipulation of 'units' attribute be done outside of this subroutine? At the place where data and attributes are added to ESMF_FieldBundle. This routine (write_netcdf) should be as generic as possible. It shouldn't "know" anything about iau_offset.

jswhit: Ideally yes, but I'm afraid that model may use this ESMF units attribute internally.  Maybe @junwang-noaa  can confirm whether it's used or not?

junwang-noaa: I looked at the code again. Write grid component actually uses IO_BASETIME ( ESMF_Time) and IO_CURRTIMEDIFF (ESMF_TimeInterval) in its internal state to specify output initial date and the forecast hours related to initial output date.  These two variable can be changed from forecast state due to various reasons, and in current code write grid component already adjusts the two time variables if iau_offset>0. Instead getting the initial output time from write grid, I think you can pass these two time variables into write_netcdf subroutine.

DusanJovic-NOAA: The "time:units" attribute is set in module_fcst_grid_comp.F90. I see iau_offset is available in that module. Can "time:units" attribute be created with correct value in that module and passed to write_netcdf routine via FieldBundle. Instead of writing incorrect value, reading it back, parsing individual time components, adjusting based on iau_offset, constructing new "hours since ..." string and finally overwriting the one already written.
Passing IO_BASETIME and IO_CURRTIMEDIFF instead of iau_offset and doing the same or similar manipulation is equally bad.

junwang-noaa: "Time:units" in fcst_grid_comp is to control the forecast integration and it is correct to provide time information from forecast point of view (starting integration from 2019103121 for 2019110100 cycle, but we decided only to take forecast results for 2019110100 history output). In my point of view "Time:units" on forecast grid comp should not be changed, because it is write grid comp decided what to output with the data. In the case of running with IAU, it's 2019110100 is the corresponding f00. That is why we use the two time variables  IO_BASETIME and IO_CURRTIMEDIFF owned by write grid comp to have consistent time stamp for all the tasks done on write grid comp(output files, inline_post etc).

DusanJovic-NOAA: The "time:units" is a string attribute added to the export state for the purpose of passing that information (and many other) to the routine that writes that state to a (netcdf) file (convention="NetCDF"). I do not see how it can "control the forecast". Where exactly is "time:units" attribute used, except in the write_netcdf routine?  If it's used then that's also not the best way of controlling the forecast. We use various esmf objects (clocks, time, time intervals and alarms) for that, not string attributes.

junwang-noaa: Sorry, I should say "time:units" on forecast grid component shows the time stamp on forecast integration, no matter what downstream routine(s) will use it, we should not change it. Write grid comp should be allowed to make decision on what to output.

jswhit: I implemented Dusan's suggestion and it seems to work.

jswhit:
The `time:units` attribute is now modified in `module_fcst_grid_comp.F90` and `module_write_netcdf.F90` is left unmodified.  Tested on hera - seems to work as expected with no unexpected side effects.

junwang-noaa:
The code is committed, close the pull request.

PR #10, opened 2019-11-19
The HAFS related developments for the write_grid_component

BinLiu-NOAA:
Bring in the HAFS related developments for the write_grid_component to 
- Enable outputting a grid (regional_latlon, rotated_latlon, or lambert_conformal) larger than the native computation grid for the regional or nested domain/tile. 
- Enable writing out the output on the native computation grid directly without any interpolation for the regional or nested domain/tile.


BinLiu-NOAA:
Notes:
- Regression tests have been done on Hera, confirming these write_grid_component changes do not change the baseline results.
- There are only code changes/modifications for three files. And if possible, please use the "squash and merge" option when doing the actual merge of this pull request.

io/module_wrt_grid_comp.F90
```diff
@@ -1749,6 +1790,269 @@ subroutine recover_fields(file_bundle,rc)
...
+           call ESMF_FieldGet(fcstField(ifld),localDe=0, farrayPtr=var2dPtr2dr4, rc=rc)
+           istart = lbound(var2dPtr2dr4,1)
+           iend   = ubound(var2dPtr2dr4,1)
+           jstart = lbound(var2dPtr2dr4,2)
+           jend   = ubound(var2dPtr2dr4,2)
+           allocate(maskwrt(istart:iend,jstart:jend))
+           maskwrt(istart:iend,jstart:jend)=1.0
+         endif
+       endif
+       if(index(trim(fieldName),"pressfc")>0) then
```
DusanJovic-NOAA, line 1854: What will happen if "pressfc" is NOT present in the file_bundle? Looks like maskwrt will remain unallocated, yet it is used in the following do loop without checking its allocation status. Have you tested this code change in such case?

DusanJovic-NOAA: I checked the code more carefully, maskwrt will be allocated but it will not be assigned value 0.0 to indicate missing value if "pressfc" is not found in the file_bundle.

BinLiu-NOAA: Hi Dusan,

Thanks for digging into the code! And yes, the assumption is that there must be a "pressfc" field in the output field bundle, otherwise the output fields will not be masked out properly for those points outside of the computation grid. But, in any case, the forecast model will still run, only with some output fields not being properly masked out. 

Appreciate it if you could provide some suggestions.

Thanks!

Bin

DusanJovic-NOAA (approved):

junwang-noaa (approved):
The code changes look good. The only comment is that the code to compute maskwrt can be done in a simpler way. Bin Liu will update the piece of code in next commit.

junwang-noaa:
This commit includes following code changes:
1. Enable outputting a grid (regional_latlon, rotated_latlon, or lambert_conformal) larger than the native computation grid for the regional or nested domain/tile.
2. Enable writing out the output on the native computation grid directly without any interpolation for the regional or nested domain/tile (VLAB issue: #61870). 
3. Change the missing value in physics bundle from real(kind_phys) to real in io/FV3GFS_io.F90 as the file is compiled in the top level and could be in 32bit real or 64bit real.

PR #11, opened 2019-11-25
Add support for cmake build system.

DusanJovic-NOAA:


RatkoVasic-NOAA:
Looks good to me. Does it work? :+1:

climbfuji (commented):
I just looked at this, didn't test it yet with the GNU compiler. The stochastic physics configuration is a little confusing, is it because you don't want to add CMakeLists.txt to the stochastic_physics repository/directory? I would also adjust the indents in the various CMakeLists.txt files (in particular the list of sources in gfsphysics/CMakeLists.txt).

Issue #13, opened 2019-11-26
update the time stamp at fh00 in inline post


junwang-noaa:
It is found that the time stamp at fh00 in inline POST has minutes in it. The following change is to fix the problem:
in io/post_gfs.F90

@@ -133,6 +133,7 @@ module post_gfs
 !
       ifhr  = mynfhr
       ifmin = mynfmin
+      if (ifhr == 0 ) ifmin = 0
       if(mype==0) print *,'bf set_postvars,ifmin=',ifmin,'ifhr=',ifhr
       call set_postvars_gfs(wrt_int_state,mpicomp,setvar_atmfile,   &
            setvar_sfcfile)

junwang-noaa:
Code is committed, close ticket.

PR #14, opened 2019-12-03
change ifmin to zero at fh00 for inline POST

junwang-noaa:
NOAA-EMC/fv3atm issue #13 : update the time stamp at fh00 in inline post


RussTreadon-NOAA (approved):
The change made to io/post_gfs.F90 is the change we have been using in the real-time GFS v16 parallel on Mars.   Output from the Mars parallel has been deemed acceptable with this change.   Given this, the proposed change should be merged.

DusanJovic-NOAA (approved):

PR #20, opened 2019-12-06
 Add support for cmake build system

DusanJovic-NOAA:


stochastic_physics/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+
+if(32BIT)
```
junwang-noaa, line 2: should this CMakeLists.txt be added to the stochastic_physics repo? I know there is dependencies on the structure of stochastic_physics, but my understanding is that this file should belong to stochastic_physics.

DusanJovic-NOAA: It should. Eventually. I just wanted to reduce the number of repositories I have to update in this first cmake commit.

junwang-noaa (approved):
Code changes look good.

PR #32, opened 2019-12-23
Bug fix in io/FV3GFS_io.F90 when using RUC LSM

climbfuji:
This PR contains a bug fix in io/FV3GFS_io.F90 when using RUC LSM for cold starts from RAP/HRRR initial conditions. The test whether to reallocate the arrays required for writing restart files needs to test different array dimensions, not just the number of soil levels. Included is also a missing if-test for NoahMP in an unrelated code.

Associated PRs: MISSING

See MISSING for regression testing information.


climbfuji:
Closing - should have been for NCAR dtc/develop.

Issue #54, opened 2020-02-04
update coupled field name in fv3


junwang-noaa:
The coupled field names in fv3 is different from that used in CICE5. In this ticket, the names will be changed so that both components will use same field names, and nems code can be clean up for name matching.

junwang-noaa:
Denise has the code changes in her branch:

https://github.com/DeniseWorthen/fv3atm/tree/feature/changefldnames

Her corresponding NEMS branch is:

https://github.com/DeniseWorthen/NEMS/tree/feature/changefldnames

junwang-noaa:
Code is committed, close issue.

PR #57, opened 2020-02-06
Change field names

DeniseWorthen:
This PR includes the following changes:
* fv3atm #54: field names imported from ice have same name as the ice export field
* PET log messages are written at the realize step to document connected and unconnected fields
* When Model%cplflx is true, land fluxes exported to the mediator for coupling to the ocean are set huge
* Adds tiled field dumping of coupled fields when statewrite_flag is true


junwang-noaa:
Code is committed in pull request [#61](https://github.com/NOAA-EMC/fv3atm/pull/61) close this pull request.

Issue #91, opened 2020-03-31
add capability to write out global regular (non-gaussian) lat/lon grid history files


jswhit:
The 'gaussian_grid' option is used for backwards compatibility with the spectral GFS, for downstream applications (like GSI) that expect gaussian latitudes.  Most applications would prefer a regular evenly-spaced latitude grid.  Therefore, it makes sense to add a 'global_latlon' option for 'output_grid' in model_configure.

jswhit2:
@aerorahul

jswhit2:
@junwang-noaa pointed out that the 'latlon_grid' option was intended to provide this functionality, but was never fully implemented.  Right now it does not produce any output.  I've gone ahead and removed the non-functional 'latlon_grid' code from module_wrt_grid_comp.F90.

jswhit2:
@yangfanglin  asked how to define the resolution of the output grid.  The parameters `imo` and `jmo` can be used for this.  For example, `imo=2880 jmo=1440` will produce a 0.125 degree grid (not including the poles and equator).  `jmo=1441` will give you a 0.125 degree grid including the poles and equator.

jswhit2:
Currently, `netcdf` and `netcdf_parallel` output is supported in PR #92.  The structure of the resulting netcdf files are identical to the gaussian grid files (except that the latitude values are different).

junwang-noaa:
The code changes were merged to develop branch in PR #104. close ticket.

PR #92, opened 2020-03-31
create 'global_latlon' output_grid option

jswhit2:
This pull request implements a 'global_latlon' option for history file output.  
It works for 'netcdf' and 'netcdf_parallel' output.
If jmo is even, the regular lat/lon grid does not include the poles or equator.
If jmo is odd, the poles and equator is included.

Addresses issue #91


edwardhartnett:
Are there any tests for this code?

jswhit2:
> Are there any tests for this code?

Not yet.

climbfuji:
This capability will be very useful. The TC trackers require the input in lat-lon format, currently we have to use an additional regridding step from the gaussian grid to a lat-lon grib using wgrib2 before running the tracker (and interpolating twice doesn't improve the accuracy).

junwang-noaa:
Code was committed in [#104](https://github.com/NOAA-EMC/fv3atm/pull/104), close PR.

PR #97, opened 2020-04-08
Initializing ice fluxes by PBL calculated values when fluxes from CIC…

ShanSunNOAA:
Initializing ice fluxes by PBL calculated values when fluxes from CICE are unavailable.


ShanSunNOAA:
Closing #97 as #97 and #107 are merged into #119.

PR #99, opened 2020-04-15
CCN-IN followup bugfix

grantfirl:
The following bug was discovered when propagating the MG CCN/IN changes to the CCPP SCM.

The variable iaerclm in the Model DDT (GFS_control_type) was conditionally initialized in GFS_typedefs.F90, leading to segmentation faults when the condition was not met (if the uninitialized variable somehow gets set to true by the compiler or an antecedent memory value).

Also, the character length of the fn_nml variable in the GFS_init_type DDT was different than other instances of the variable in the file. Changed the character length to 64 from 65 to be consistent. Credit to @pjpegion for finding this one.

In addition, two other minor bugfixes are included:

Model%debug is now initialized before use in GFS_typedefs.F90 and a redundant code block initializing variables for NoahMP was deleted.

This PR was initiated on the request of @climbfuji following successful CCPP SCM testing.

### DIFF
--- a/gfsphysics/GFS_layer/GFS_typedefs.F90
+++ b/gfsphysics/GFS_layer/GFS_typedefs.F90
@@ -147,7 +147,7 @@ module GFS_typedefs
 
     character(len=32), pointer :: tracer_names(:) !< tracers names to dereference tracer id
                                                   !< based on name location in array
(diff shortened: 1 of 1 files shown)


junwang-noaa:
Code is closed in [#104](https://github.com/NOAA-EMC/fv3atm/pull/104), close PR.

PR #107, opened 2020-04-27
Add a separate array for internal ice temperature

ShanSunNOAA:
Sfcprop%stc is currently used to store temperatures for both soil and internal ice. In the case of fractional grid, both surface types can exist at a single grid point. Thus we introduced a separate array called "tiice" to store internal ice temperature. "tiice" will be added to phyf*nc output only when frac_grid=T.


ShanSunNOAA:
Closing #97 as #97 and #107 are merged into #119.

Issue #142, opened 2020-07-10
Initializing non-zero ice fluxes in the cold coupled run with CMEP


ShanSunNOAA:
Currently dusfci_cpl, dvsfci_cpl, dtsfci_cpl and dqsfci_cpl are initialized by values from PBL in the cold run at the 1st time step, which happens to be flag_cice=F for NEMS. However, "flag_cice" is not the right flag for "1st time cold run" in the case of CMEPS. We replaced the flag_cice condition by dusfcin_cpl, i.e., when ice flux is huge (undefined), use PBL-calculated ice fluxes instead. This will change results in the coupled model only. Passed regression tests for standalone FV3.

This needs to go with ccpp-physics #472.

climbfuji:
@shansun6 will this be fixed by https://github.com/NCAR/ccpp-physics/pull/563 ?

ShanSunNOAA:
Yes it is the same thing but at a higher FV3atm level. My misunderstanding
for creating an issue there as well.

Thanks,
Shan

climbfuji:
This will be fixed in https://github.com/NCAR/ccpp-physics/pull/563.

PR #143, opened 2020-07-14
Initialize ice fluxes to be huge

ShanSunNOAA:
When ice flux is huge (undefined in CICE yet) in the 1st time step cold run, use PBL-calculated ice fluxes instead. This will change results in the coupled model only. Passed regression tests for standalone FV3.

This refers to FV3 issue #142 and CCPP-physics issue #472.

@junwang-noaa @SMoorthi-emc

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,5 +8,5 @@
 	branch = master
 [submodule "ccpp/physics"]
 	path = ccpp/physics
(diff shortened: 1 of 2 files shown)


junwang-noaa:
The associated ccpp PR 472 was closed, so this PR is closed.

PR #145, opened 2020-07-21
Make stochastic physics independent of fv3atm and its submodules

climbfuji:
This PR and associated PRs make the stochastic_physics submodule independent of fv3atm and its submodules.

To reduce the complexity in `atmos_model.F90`, the stochastic physics / cellular automata calls where moved into a `stochastic_physics_wrapper` module in the existing subdirectory `stochastic_physics`. Going into the future, I hope we can remove the entire blocked data structure construct from the actual stochastic physics / cellular automata code and then simplify the contents of the `stochastic_physics_wrapper` module.

Also included: PR and issue templates for fv3atm (from @junwang-noaa).

Associated PRs:

https://github.com/noaa-psd/stochastic_physics/pull/27
https://github.com/NOAA-EMC/fv3atm/pull/145
https://github.com/ufs-community/ufs-weather-model/pull/171

For regression testing information, see https://github.com/ufs-community/ufs-weather-model/pull/171.


pjpegion (commented):
I see you put a stochastic_physics_wrapper in FV3/stochastic_physics. I thought this directory was only for the cmake build, and was supposed to go away?

climbfuji:
> I see you put a stochastic_physics_wrapper in FV3/stochastic_physics. I thought this directory was only for the cmake build, and was supposed to go away?

I could have placed the entire code in this wrapper in `atmos_model.F90`, either inline where it was beforehand or in separate subroutines. The issue is that the code is nearly identical for both the `init` and `run` phases, and that it was largely duplicated before I made the changes. Now we have even more duplicated code if we leave the calls in the `init` and `run` phases in `atmos_model.F90`, thus outsourcing it into separate subroutines - or for clarity its own module - avoids duplication. It also allows us to make changes to stochastic physics later (remove the blocking etc) and then only change the wrapper. To be clear, the wrapper is _independent_ of the actual stochastic_physics package and a part of fv3atm. Whether it sits inside a subdirectory `stochastic_physics` or next to `atmos_model.F90` or even in `atmos_model.F90` as additional subroutines doesn't matter. I don't like that latter, though, because `atmos_model.F90` is already long and messy.

pjpegion:
I will leave that decision up to you and others.  I did some testing, and these changes are working as expected.

climbfuji:
> I will leave that decision up to you and others. I did some testing, and these changes are working as expected.

Great, thanks. I will wait for this decision before fixing the gnumake build.

stochastic_physics/makefile
```diff
@@ -0,0 +1,83 @@
...
+OBJS_c   = $(SRCS_c:.c=.o)
+
+OBJS = $(OBJS_f) $(OBJS_f90) $(OBJS_F) $(OBJS_F90) $(OBJS_c)
+
+all default: depend $(LIBRARY)
+
+$(LIBRARY): $(OBJS)
+	$(AR) $(ARFLAGS) $@ $?
+
+# FV3GFS_io.o: FV3GFS_io.F90
```
DusanJovic-NOAA: The following ~20 lines can be deleted.

climbfuji: Done.

pjpegion (approved):
Changes look good.

DusanJovic-NOAA (approved):

PR #153, opened 2020-08-04
Add dependencies to CCPP metadata

climbfuji:
This PR removes dependency information from the CCPP prebuild config. See https://github.com/NCAR/ccpp-framework/issues/308 and https://github.com/NCAR/ccpp-framework/pull/317 for details on the motivation for this change and the actual implementation. It also removes some legacy code used by the dynamic CCPP build in the past.

This PR also contains the changes in https://github.com/NOAA-EMC/fv3atm/pull/156, i.e. the completion of adding the `active` attribute to `GFS_typedefs.F90`. On top of this PR, the missing `active` attribute for `phy_fctd` is added.

Additionally, gfortran-10 compiler flags are added to CCPP's `CMakeLists.txt`.

Associated PRs:
https://github.com/earth-system-radiation/rte-rrtmgp/pull/85
https://github.com/NCAR/ccpp-framework/pull/317
https://github.com/NCAR/ccpp-physics/pull/483
https://github.com/NOAA-EMC/fv3atm/pull/153
https://github.com/ufs-community/ufs-weather-model/pull/180

See https://github.com/ufs-community/ufs-weather-model/pull/180 for regression testing.


gfsphysics/GFS_layer/GFS_typedefs.meta
```diff
@@ -680,6 +690,7 @@
   dimensions = (horizontal_dimension)
   type = real
   kind = kind_phys
+  active = (flag_for_mass_flux_deep_convection_scheme == flag_for_gf_deep_convection_scheme)
```
junwang-noaa, line 693: I am curious how the active is used? Will the variable be allocated if active is false or this attribute will only be used by the ccpp/framework?

junwang-noaa (approved):

gfsphysics/GFS_layer/GFS_typedefs.meta
climbfuji: The attribute is currently only used by the ccpp-framework to perform or skip certain operations in the auto-generated caps. In the future, data entirely internal to physics (i.e. everything in the GFS_interstitital DDT) will be allocated automatically by the ccpp-framework, and then `active` can be used to skip certain allocations.

But for the original/other GFS DDTs in `GFS_typedefs.F90`, it will only be used in the auto-generated caps. That's why the Fortran allocation logic in `GFS_typedefs.F90` needs to be consistent with the `active` attribute in `GFS_typedefs.meta` for those data.

DusanJovic-NOAA (approved):

PR #154, opened 2020-08-06
Cldfra

ericaligo-NOAA:
## Description

Diagnostic cldfra array outputs 3D instantaneous cloud fractions.  Works with WSM6, Thompson and GFDL mp schemes.  No results changed.

UPDATE:  Provided a bug fix to GFS_restart.F90 with an if condition added in the ifdef CCPP block to test for reflectivity flag. RTs have been updated. Test run in 3-km FV3 LAM was successful and plots looked good.

### Issue(s) addressed

https://github.com/ufs-community/ufs-weather-model/issues/158
https://github.com/NOAA-EMC/fv3atm/pull/154
https://github.com/NCAR/ccpp-physics/pull/484

## Testing

Tested on hera with the SAR configuration.  Evaluated fields from run with Thompson microphyics and GFDL microphysics. 
Full RTs performed on hera.

## Dependencies
FV3
ccpp/physics

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,5 +8,5 @@
 	branch = master
 [submodule "ccpp/physics"]
 	path = ccpp/physics
-	url = https://github.com/NCAR/ccpp-physics
-	branch = master
+	url = https://github.com/ericaligo-NOAA/ccpp-physics.git 
+	branch = cldfra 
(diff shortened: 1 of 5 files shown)


junwang-noaa (approved):
If I understand correctly, the cldfra is an optional diagnostic field, I'd hope to have a logic variable to control if to allocate the whole 3D array as other 3D diagnostic tendency fields does, currently this array will be allocated and computed for all the applications even though it is not required to output. This can be done in a later commit.

ChunxiZhang-NOAA (approved):
I didn't see any problem.

Issue #174, opened 2020-09-22
add parallel netcdf capability to regional fv3


junwang-noaa:
## Description
Regional fv3 needs the parallel writing compressed netcdf data capability.

## Solution
Add the parallel writing compressed netcdf data capability for regional fv3 grid. 

## Testing:

It will be tested for LAM and HAFS high resolution runs. 
RT will be conducted on supported platforms, the current baseline results will not change.  

## Dependent PRs:

PRs will be submitted in FV3, NEMS and ufs-weather-model.

junwang-noaa:
Committed, close issue.

PR #175, opened 2020-09-22
add parallel netcdf for regional applications

junwang-noaa:
## Description of code changes:

1. add parallel netcdf call for regional fv3 grids
2. Change restart_interval to real from integer so that model can output restart files at non-integer forecast time, e.g.  output restart files after the first time step.
3. Set forecast time to be real to allow flexible forecast time. 

### Issue(s) addressed

[FV3 issue #174](https://github.com/NOAA-EMC/fv3atm/issues/174)

## Testing

The code will be tested for LAM and HAFS high resolution runs.
RT will be conducted on supported platforms, the current baseline results will not change.

## Dependencies

[fv3 PR #175](https://github.com/NOAA-EMC/fv3atm/pull/175)
[NEMS PR #81](https://github.com/NOAA-EMC/NEMS/pull/81)
[ufs-weather-model PR#206](https://github.com/ufs-community/ufs-weather-model/pull/206)


BinLiu-NOAA (approved):
Changes look good to me. Tests were previously conducted with the UFS HAFS application for the parallel netcdf output and for the flexible restart and model forecast times.

DusanJovic-NOAA (approved):

JamesAbeles-NOAA (approved):
The code does not allow for compression with parallel_netcdf but does for netcdf. I approve the change but hope we can revisit this so that compression is an option if the output grid does not contained iundefined points or if there is a way to compress the data when the output points that are undefined are set to something netcdf knows are undefined and doesn't casue compression errors

junwang-noaa:
Jim, to clarify, the code does allow the lossless compression for both parallel_netcdf and serial netcdf. As you mentioned we can revisit lossy compression in the future.  Thanks

PR #200, opened 2020-11-20
Bug fixes in (1) running with frac_grid=T and GFDL MP and (2) restarting with frac_grid=T

ShanSunNOAA:
The crash in the frac_grid=T and GFDL MP run is fixed by using GFS_surface_composites.F90/meta in Moorthi's branch "SM_Oct102020".

Restart can reproduce after
-- Not setting fice to zero in order to leave lake ice untouched in atmos_model.F90
-- Bad water temperature is only filtered at initial time in FV3GFS_io.F90

Co-authored-with: Shrinivas Moorthi <shrinivas.moorthi@noaa.gov>
Co-authored-with: Denise Worthen <Denise.Worthen@noaa.gov>

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,5 +8,5 @@
 	branch = master
 [submodule "ccpp/physics"]
 	path = ccpp/physics
-	url = https://github.com/NCAR/ccpp-physics
-	branch = master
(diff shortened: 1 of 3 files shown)


ShanSunNOAA:
This is replaced by #204.

PR #202, opened 2020-11-21
Update develop from gsd/develop 2020/11/20: Unified gravity wave drag, updates to other GSL physics

### LINKED ISSUE #187: Pretty print compile output when calling ccpp_prebuild.py
## Description

A recent change moved the call to ccpp_prebuild.py from ufs-weather-model to fv3atm. Since then, the output in the build log reads like
```
Calling CCPP code generator (ccpp_prebuild.py) ... --suites=FV3_HAFS_ferhires,FV3_HAFS_FA_HWRF_RRTMG
```
when the argument `--suites` is specified. It used to be - and should be again -
```
Calling CCPP code generator (ccpp_prebuild.py) for suites XXX ... 
```
where XXX is the comma-separated list of suites, i.e. `FV3_HAFS_ferhires,FV3_HAFS_FA_HWRF_RRTMG`. Alternatively, it could be 
```
Calling CCPP code generator (ccpp_prebuild.py) with argument --suites=XXX ... 
```
If no suites area specified it should be (and I think it does)
```
Calling CCPP code generator (ccpp_prebuild.py) ... 
```

climbfuji:
## Description

This PR updates the GSL physics and does a few other things:
- add `--debug` to `ccpp_prebuild.py` call when model is compiled in DEBUG mode, and pretty print the output in the build log (fixes #187 and https://github.com/ufs-community/ufs-weather-model/issues/135)
- add unified gravity wave drag (called unified ugwp) that combines the UGWP v1 with the GSL drag suite
- add suite definition file for GFS v16beta for coupled model
- move static array `zs` (depth of soil layer), dimensioned as `1:lsoil_lsm` from `Sfcprop` to `Model` (since this is a model configuration and only needs to be stored once, not for each block)

Note. This PR needs to be squashed at merge time, because it the commit history shows all the individual commits from the previous merges (which were squashed as well). Following this commit, the gsd/develop branches in the NOAA-GSL fork will be deleted and we will start with a clean plate from develop for our new branches gsl/develop.

### Issue(s) addressed

https://github.com/ufs-community/ufs-weather-model/issues/135
#187 

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/297

## Dependencies

https://github.com/NCAR/ccpp-framework/pull/338
https://github.com/NCAR/ccpp-physics/pull/521
https://github.com/NOAA-EMC/fv3atm/pull/202
https://github.com/ufs-community/ufs-weather-model/pull/297


climbfuji:
@SMoorthi-emc @DeniseWorthen @JessicaMeixner-NOAA before I open this PR for review, can you please check commit https://github.com/NOAA-EMC/fv3atm/pull/202/commits/c2890282d88fc65e81ff41efb03fe7e660c93d8f, which adds the GFSv16beta coupled suite? Thank you!

SMoorthi-emc:
Hi Dom,
   It looks like a coupled suite with GFDL microphysics.  Do we need to review before adding new suites now?
Adding new suites doesn't cost anything and since since different physics are bases on suites, we need to have the flexibility to add new suites (for example I have a version of this suite with RAS, in place of SAS.
Thanks
Moorthi

climbfuji:
I just want to make sure that the suite I am supposed to add for the coupled model is ok (I was asked to add this suite as part of my PR). Thanks for taking a look!

JessicaMeixner-NOAA:
@DomHeinzeller thanks for adding the new suite for coupling!

CMakeLists.txt
```diff
@@ -1,9 +1,14 @@
 if(CCPP)
 
+  if(DEBUG)
+    set(_ccpp_debug_arg "--debug")
```
aerorahul, line 4: where does this get passed to `ccpp_prebuild.py`?

climbfuji: That's embarrassing. I fixed it, thanks.

aerorahul (approved):
I don't have any expertise in evaluating the scientific content of this PR, so I will defer to the other reviewers.

DusanJovic-NOAA (approved):

ccpp/suites/suite_FV3_GSD_noah.xml
```diff
@@ -79,6 +79,7 @@
       <scheme>mp_thompson_post</scheme>
       <scheme>GFS_MP_generic_post</scheme>
       <scheme>cu_gf_driver_post</scheme>
+      <!-- <scheme>tracer_sanitizer</scheme> -->
```
junwang-noaa, line 82: Is this  a place holder? tracer_sanitizer seems not used in any suite file?

climbfuji: That's right, the "offending" `tracer_sanitizer` scheme is added in ccpp-physics for testing and debugging stability issues. It's not to be used ever in applications, because it's a dumb scheme that simply resets negative tracers to zero and doesn't care about water conservation. As such, it is commented out everywhere, but the commented lines for the two GSD suites illustrate how we used it in order to stabilize the model. It's similar to the Thompson MP tendency limiter, which simply puts a lid on the heating rate coming out of the Thompson scheme without adjusting any other quantity. I guess we will delete the `tracer_sanitizer` and Thompson MP tendency limiter once we figured out how to run the GSL physics in a stable way at reasonable time steps (it's always stable at small timesteps, but these can't be used in operations).

io/FV3GFS_io.F90
```diff
@@ -703,17 +704,24 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain)
...
-          Sfcprop(nb)%hprime(ix,5)  = oro_ls_var(i,j,5)
-          Sfcprop(nb)%hprime(ix,6)  = oro_ls_var(i,j,6)
-          Sfcprop(nb)%hprime(ix,7)  = oro_ls_var(i,j,7)
-          Sfcprop(nb)%hprime(ix,8)  = oro_ls_var(i,j,8)
-          Sfcprop(nb)%hprime(ix,9)  = oro_ls_var(i,j,9)
-          Sfcprop(nb)%hprime(ix,10)  = oro_ls_var(i,j,10)
+          ! Replace hprime(1:10) with GSL oro stat data only when using GSL
+          ! drag suite with large scale GWD and blocking as part of unified drag
+          ! suite. Otherwise, original oro stat data is used.
+          if ( (Model%gwd_opt==3 .or. Model%gwd_opt==33) .or.    &
```
junwang-noaa, line 710: What about Model%gwd_opt==1? So it seems previous code will update hprime for all the gwd options, with the if statement, hprime will only be updated for options 3,33 or certain conditions with gsl drag.

climbfuji: `gwd_opt = 1` refers to the original GFS GWD (in `gwdps.f`). For this, the separate large scale and small scale orography files are not in use. These were created specifically for the unified gravity wave drag, but they are only used to replace the default orography (that is read previously) under certain conditions. For gwd_opt = 2 or 22 (the "new" unified gravity wave drag from this PR, with (22) or without (2) extended diagnostics turned on), the new large scale orography files are only used if the flag for "large-scale GWD and blocking only" = do_gsl_drag_ls_bl is on. They are always used for gwd_opt 3 or 33 (the GSL drag suite). @mdtoyNOAA did I capture this correctly?

climbfuji:
This is ready to merge anytime.

junwang-noaa (approved):

io/FV3GFS_io.F90
mdtoy: Dom.  Yes, this is correct.

SMoorthi-emc (approved):

PR #204, opened 2020-11-30
Bug fixes in (1) running with frac_grid=T and GFDL MP and (2) restarting with frac_grid=T

ShanSunNOAA:
The crash in running GFDL MP with frac_grid=T is fixed by using GFS_surface_composites.F90/meta in Moorthi's branch "SM_Oct102020".

Restart can reproduce after
-- Not setting fice to zero in order to leave lake ice untouched in atmos_model.F90
-- Bad water temperature is only filtered at initial time in FV3GFS_io.F90
-- Moorthi's CCPP update

Co-authored-with: Shrinivas Moorthi shrinivas.moorthi@noaa.gov
Co-authored-with: Denise Worthen Denise.Worthen@noaa.gov

This replaced PR #200

### DIFF
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -1764,7 +1764,6 @@ end subroutine atmos_data_type_chksum
                   nb = Atm_block%blkno(i,j)
                   ix = Atm_block%ixp(i,j)
 
-                  IPD_Data(nb)%Sfcprop%fice(ix)          = zero
                   IPD_Data(nb)%Coupling%slimskin_cpl(ix) = IPD_Data(nb)%Sfcprop%slmsk(ix)
                   ofrac = IPD_Data(nb)%Sfcprop%oceanfrac(ix)
                   if (ofrac > zero) then
@@ -1779,7 +1778,7 @@ end subroutine atmos_data_type_chksum
                       if (abs(one-ofrac) < epsln) then
                         IPD_Data(nb)%Sfcprop%slmsk(ix)         = zero
                         IPD_Data(nb)%Coupling%slimskin_cpl(ix) = zero
-                      end if
+                      endif
                     endif
                   endif
                 enddo
--- a/io/FV3GFS_io.F90
+++ b/io/FV3GFS_io.F90
@@ -1066,14 +1066,7 @@ module FV3GFS_io_mod
         endif
 
         if(Model%frac_grid) then ! obtain slmsk from landfrac
-!! next 5 lines are temporary till lake model is available
-          if (Sfcprop(nb)%lakefrac(ix) > zero) then
-!           Sfcprop(nb)%lakefrac(ix) = nint(Sfcprop(nb)%lakefrac(ix))
-            Sfcprop(nb)%landfrac(ix) = one - Sfcprop(nb)%lakefrac(ix)
-            if (Sfcprop(nb)%lakefrac(ix) == zero) Sfcprop(nb)%fice(ix) = zero
-          endif 
-          Sfcprop(nb)%slmsk(ix) = ceiling(Sfcprop(nb)%landfrac(ix))
-          if (Sfcprop(nb)%fice(ix) > Model%min_lakeice .and. Sfcprop(nb)%landfrac(ix) == zero) Sfcprop(nb)%slmsk(ix) = 2 ! land dominates ice if co-exist
+          Sfcprop(nb)%slmsk(ix) = ceiling(Sfcprop(nb)%landfrac(ix)) !nint/floor are options
         else ! obtain landfrac from slmsk
           if (Sfcprop(nb)%slmsk(ix) > 1.9_r8) then
             Sfcprop(nb)%landfrac(ix) = zero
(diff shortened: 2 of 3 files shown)


climbfuji:
Associated PRs:

https://github.com/NCAR/ccpp-physics/pull/524
https://github.com/NOAA-EMC/fv3atm/pull/204
https://github.com/ufs-community/ufs-weather-model/pull/304

For regression testing information, see https://github.com/ufs-community/ufs-weather-model/pull/304.

SMoorthi-emc (approved):

atmos_model.F90
```diff
@@ -1764,7 +1764,6 @@ subroutine assign_importdata(rc)
                   nb = Atm_block%blkno(i,j)
                   ix = Atm_block%ixp(i,j)
 
-                  IPD_Data(nb)%Sfcprop%fice(ix)          = zero
```
junwang-noaa, line 1767: The line may set lake ice to be zero, can we confirm the impact of removing this line on lake ice?

io/FV3GFS_io.F90
```diff
@@ -1084,16 +1077,32 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain)
...
         if (Sfcprop(nb)%lakefrac(ix) > zero) then
           Sfcprop(nb)%oceanfrac(ix) = zero ! lake & ocean don't coexist in a cell
-!         if (Sfcprop(nb)%fice(ix) < Model%min_lakeice) then
-!            Sfcprop(nb)%fice(ix) = zero
-!            if (Sfcprop(nb)%slmsk(ix) == 2) Sfcprop(nb)%slmsk(ix) = 0
-!         endif
+          if (Sfcprop(nb)%slmsk(ix) /= one) then
+            if (Sfcprop(nb)%fice(ix) >= Model%min_lakeice) then
+              if (Sfcprop(nb)%slmsk(ix) < 1.9_r8)      &
+                write(*,'(a,2i3,3f6.2)') 'reset lake slmsk=2 at nb,ix=' &
```
junwang-noaa, line 1083: I am curious how many print lines we will get from the write statement here? Do we need them?

io/FV3GFS_io.F90
```diff
@@ -1351,7 +1360,7 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain)
 !$omp parallel do default(shared) private(nb, ix, tem, tem1)
       do nb = 1, Atm_block%nblks
         do ix = 1, Atm_block%blksz(nb)
-          Sfcprop(nb)%tsfco(ix) = max(con_tice, Sfcprop(nb)%tsfco(ix))
+          if( Model%phour < 1.e-7) Sfcprop(nb)%tsfco(ix) = max(con_tice, Sfcprop(nb)%tsfco(ix)) ! this may break restart reproducibility
```
junwang-noaa, line 1363: May I ask why we need this if statement condition?

io/FV3GFS_io.F90
SMoorthi-emc: Yes, I have confirmed in my own experiments that ice fractionas over lakes are maintained after this fix.
The motivation that lead to finding this bug was to understand why ice concentration over lakes vanished in the coupled runs while the initial conditon had non zero values.

atmos_model.F90
SMoorthi-emc: Yes, I have confirmed in my own experiments that ice fractionas over lakes are maintained after this fix.
The motivation that lead to finding this bug was to understand why ice concentration over lakes vanished in the coupled runs while the initial condition had non zero values

io/FV3GFS_io.F90
SMoorthi-emc: Wrong place, please ignore the above comment.

io/FV3GFS_io.F90
SMoorthi-emc: Because this will fix if the initial conditon contains unphysical values.

ShanSunNOAA:
Hi Jun, this one is ok, as it is inside if
(IPD_Data(nb)%Sfcprop%oceanfrac(ix)>0.).  There was one line setting fice
to zero without any condition, which Moorthi had removed.

Thanks for checking,
-Shan

climbfuji (commented):
Looks good to me. Will approve once ccpp-physics is merged, the submodule pointer is updated and the `.gitmodules` change is reversed.

io/FV3GFS_io.F90
junwang-noaa: We may need to consider commenting these print lines in future commit as there might be too many lines in the log file.

junwang-noaa (approved):

DusanJovic-NOAA (approved):

ShanSunNOAA:
Hi Jun,

Should I comment out these lines in my upcoming commit today?

Shan

junwang-noaa:
We can do it in future commit, if you make changes in the code now, you may
have to run RT again.

ShanSunNOAA:
Ok, I will commit as is.   -Shan

climbfuji (approved):

PR #247, opened 2021-02-16
Hotfix for RRTMGP using multiple threads

dustinswales:
This PR contains several changes to the initialization and setup of the RRTMGP radiation scheme to allow for use across multiple openMP threads.
*) Moved Interstitial RRTMGP DDTs (ty_rrtmgp_gas_optics, ty_cloud_optics) to static fields defined during initialization in module memory.
*) Add "$omp critical" statements around the calling type-bound procedures during initialization.
*) Move allocation of ty_gas_conc from Interstitial to GFS_typedefs. Initialize ty_gas_concs for all blocks during initialization.

This issue was brought to our attention by @yangfanglin and Qingfu Liu at EMC while testing RRTMGP in high-resolution cases, which require multiple threads.

The GP regression tests on hera using Intel were successful. 
There are no changes to the baselines

Waiting on https://github.com/NCAR/ccpp-physics/pull/568

### DIFF
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1998,6 +1998,7 @@ module GFS_typedefs
 
     ! RRTMGP
     real (kind=kind_phys)               :: minGPpres                            !< Minimum pressure allowed by RRTMGP.
+    real (kind=kind_phys)               :: minGPtemp                            !< Minimum temperature allowed by RRTMGP. 
     integer                             :: ipsdlw0                              !<
     integer                             :: ipsdsw0                              !<
     real (kind=kind_phys), pointer      :: p_lay(:,:)                => null()  !<
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1997,6 +1997,7 @@ module GFS_typedefs
     real (kind=kind_phys), pointer      :: dudt_tms(:,:)      => null()  !< daily aver u-wind tend due to TMS
 
     ! RRTMGP
+    real (kind=kind_phys)               :: minGPpres                            !< Minimum pressure allowed by RRTMGP.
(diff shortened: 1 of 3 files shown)


yangfanglin:
Please add Qinfu Liu as a reviewer

junwang-noaa (approved):

Qingfu-Liu:
Dusan,
    Thank you very much. I will start reviewing the code changes.
Qingfu

Qingfu-Liu (approved):

DusanJovic-NOAA (approved):

ccpp/data/GFS_typedefs.F90
```diff
@@ -6438,6 +6437,14 @@ subroutine interstitial_create (Interstitial, IM, Model)
        allocate (Interstitial%toa_src_sw           (IM,Model%rrtmgp_nGptsSW))
        allocate (Interstitial%toa_src_lw           (IM,Model%rrtmgp_nGptsLW))
        allocate (Interstitial%active_gases_array   (Model%nGases))
+       ! ty_gas_concs
```
climbfuji, line 6440: Obviously %ncol and %nlay are fixed throughout the run.

What about gas_name, concs, conc? Do they depend on each grid, and if yes, do they need to be persistent? If not, then we should add them to the %rad_reset calls for the DDT (for clarity) - in the next PR. If they need to be persistent, they need to be moved to radtend or something like that (also in the next PR).

climbfuji (approved):

ccpp/data/GFS_typedefs.F90
dustinswales: @climbfuji 
I set the ncol and nlay here once in the DDT, type-bound procedures rely on it later.
gas_concentrations(ty_gas_concs) should be added to rad_reset. I will make sure to include this in my next PR.

PR #241, opened 2021-01-26
UGWP v0 v1 combined

climbfuji:
## Description

This PR
- combines the changes in PRs #227 and #236
- updates the code to the top of fv3atm develop
- updates the standard names introduced by UGWP v1 (in collaboration with the DTC CCPP physics developers)
- consolidates the diagnostic variables used in the GSL drag suite, UGWP v0 and UGWP v1
### Issue(s) addressed

n/a

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/396

## Dependencies

https://github.com/NCAR/ccpp-physics/pull/561
https://github.com/NOAA-EMC/fv3atm/pull/241
https://github.com/ufs-community/ufs-weather-model/pull/396

Issue #214, opened 2020-12-15
Remove IPD


climbfuji:
## Description

Now that we have a GFS v16 reference tag for the ufs-weather-model develop branch and its submodules, it is time to remove IPD.

## Solution

This should be done in several steps.

1. Remove all source files that are required for IPD only (in fv3atm) and make `CCPP=Y` the default in the ufs-weather-model compile/build scripts. This should not change the regression test results. However, since `CCPP=Y` is no longer in the MAKEOPT string, the suffices `_prod` and `_ccpp` for the regression test baseline/run directories will disappear (hence the requirement to create new baselines).

2. Move the computation of snow cover and other variables from `io/FV3GFS_io.F90` into CCPP (probably  the `time_vary` group), and remove the `soilveg` initialization in `GFS_driver.F90`. This should not change the regression test results when the fv3atm and the dycore are compiled in double precision (64bit), but it will change the results when fv3atm and the dycore are compiled in single precision (32bit). This is because the computation fo snow cover and other variables in `FV3GFS_io.F90` uses the same precision as fv3atm and the dycore, whereas it is always double precision in the CCPP physics.

3. Remove all `#ifndef CCPP` and related preprocessor directives from fv3atm and the ufs-weather-model, but not GFDL_atmos_cubed_sphere (These will still be needed for compatibility with how GFDL is running physics? To be confirmed with @bensonr). Remove definition of IPD data types in fv3atm and replace use of IPD data types with GFS data types. This should not change the regression test results. Update: the preprocessor directives can also be removed from the dycore.

4. Rename all regression tests (drop `_ccpp` from `fv3_ccpp_xyz`). All non-CCPP `*_run.IN` templates in `tests/fv3_conf` should be removed, and all CCPP templates renamed (drop `ccpp_` from `ccpp_xyz_run.IN`). Drop suffices `_ccpp`/`_prod` from regression test baseline/run directories (but need to keep repro? add `_debug` in the same way? or manually change the regression test control directories for `REPRO=Y` to `_repro`?)

5. Clean up all `CMakeLists.txt` files that contain legacy code to support building IPD and CCPP, using a top-level cmake or gnumake build. Now we only need to support CCPP with a cmake build.

6. Anything that is missing in the previous bullets.

## Alternatives

No alternative.

## Testing:

Testing will be done on all tier-1,2,3 platforms (tier-1 pre-commit, tier-2,3 as time permits).

## Dependent PRs:

[...]

PR #249, opened 2021-02-21
Move Noah MP init to CCPP and add new variables for Noah MP

### LINKED ISSUE #206: Duplicate symbols on macOS with Intel
I'm getting duplicate symbols on macOS building with Intel 2021.1

I encountered this bug (different files) a few months ago which was fixed with https://github.com/NOAA-EMC/fv3atm/pull/173 (`gfsphysics/CMakeLists.txt`)

The problem is that `noahmp_tables.f90`, `CCPP_data.F90`, and `namelist_soilveg.f` are being built twice. Once in `CCPP_SOURCES` and once in `IPD_SOURCES` and then their symbols conflict when linked at the end. 

https://github.com/NOAA-EMC/fv3atm/blob/a898cca15f8c3bdcc2c6d30de1e4bbe1ed9da110/gfsphysics/CMakeLists.txt#L19

This issue reared its head because of a change I had to make in the weather model's CMakeLists because I was getting undefined symbols with Intel + macOS.

I had to add this line to fix the undefined symbols, but as a side-effect the symbols in the aforementioned files now conflict: 

```
if(APPLE)
  # The linker on macOS does not include `common symbols` by default
  # Passing the -c flag includes them and fixes an error with undefined symbols
  set(CMAKE_Fortran_ARCHIVE_FINISH "<CMAKE_RANLIB> -c <TARGET>")
endif()
```

See: https://github.com/ufs-community/ufs-weather-model/issues/309

This shows up with Intel and not GCC because they handle uninitialized module variables differently. Intel marks them as `C` (common) symbols (which aren't included by default by macOS's linker), but GCC marks them as `S` (uninitialized data section for small objects).

climbfuji:
## Description

This PR does several things:

1. Move the Noah MP initialization of surface fields from fv3atm (`io/FV3GFS_io.F90`) to CCPP from @climbfuji.
2. Updates to Noah MP and Noah MP - radiation coupling from @helinwei-noaa. The changes in ccpp-physics require adding four additional variables in `GFS_typedefs.F90`.
3. Revert a change in `ccpp/data/GFS_typedefs.F90`, made as part of the UGWP update PR (https://github.com/NOAA-EMC/fv3atm/pull/241), that permanently disabled the old GFS v12-v15 convective gravity wave drag scheme. @SMoorthi-emc @ValeryYudin-NOAA please check, around line 4030 in the updated `GFS_typedefs.F90`.
4. Cleanup work in `GFS_typdefs.F90` for the recently merged RRTMGP updates: Add `Interstitial%gas_concentrations%concs(:)%conc` to the `Interstitial%rad_reset` call, remove RRTMGP variables that no longer exist from `GFS_typedefs.meta`, add missing active flags to RRTMGP variables in `GFS_typedefs.meta`

### Issue(s) addressed

This finally completes step 2 of https://github.com/NOAA-EMC/fv3atm/issues/214.

Fixes https://github.com/NOAA-EMC/fv3atm/issues/206.

## Testing

For regression testing, see https://github.com/ufs-community/ufs-weather-model/pull/425.

## Dependencies

https://github.com/NCAR/ccpp-physics/pull/575
https://github.com/NOAA-EMC/fv3atm/pull/249
https://github.com/ufs-community/ufs-weather-model/pull/425


CMakeLists.txt
```diff
@@ -109,7 +108,9 @@ target_include_directories(fv3dycore INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT
                                                $<INSTALL_INTERFACE:mod>)
 
 target_link_libraries(fv3dycore PUBLIC fms
-                                       gfsphysics
```
junwang-noaa, line 112: I am curious, we only had noahmp_tables.f90 in gfsphysics directory, does fv3dycore have dependency on that file? Can we remove this line?

DusanJovic-NOAA (approved):

CMakeLists.txt
climbfuji: The dycore has dependencies on CCPP because of the fast physics calls (saturation adjustment) in `fv_mapz.F90`. It won't compile w/o these dependencies. The `gfsphysics` line had the CCPP dependencies included implicitly because of line 126 below in the old code.

junwang-noaa: So fv3dycore has no dependency on gfsphysics itself, previous code lists gfsphysics not ccpp as dependency, maybe ccpp physics can be listed here explicitly? Also do we need ccpp? It seems the dependency is on ccppdata and ccppphys.

climbfuji: The new code has dependencies ccppdata (for the data structures), ccppphys for the fast physics code, and ccpp (=ccpp-framework) for the CCPP API:
```
target_link_libraries(fv3dycore PUBLIC fms
                                       ccpp
                                       ccppphys
                                       ccppdata
                                       esmf)
```

junwang-noaa: Sorry I don't get the dependency of ccpp (ccpp-framwork). I run the GFSv16 test without the line of ccpp, the code runs fine. 
 target_link_libraries(fv3dycore PUBLIC fms
 \-                                       ccpp
                                        ccppphys
                                        ccppdata
                                        esmf)

climbfuji: I can try. Maybe it's yet another implicit dependency in ccppphys.

barlage (approved):

CMakeLists.txt
climbfuji: It works without that line, too. But I am wondering if it is the right thing to do instead of spelling out all dependencies explicitly. @DusanJovic-NOAA what do you think?

DusanJovic-NOAA: If `fv3dycore` does not depend directly on `ccpp` (ie. ccpp-framwork), then I would leave it out. BTW at some point we should rename it from `ccpp` to `ccppframework` to be more explicit about what it is.

climbfuji: The dycore does depend explicitly on ccpp=ccppframework, because it uses `call ccpp_physics_init` etc.

DusanJovic-NOAA: In that case, specify `ccpp` as explicit dependency.

climbfuji: Ok, thanks.

junwang-noaa:
In the compile log file, the ccpp_static_api.F90 with ccpp_physics_init and
ccpp_physics_run are build in  FV3/ccpp/driver/libccppdriver.a, not in
FV3/ccpp/framework/src/libccpp.a. So it seems not from ccpp framework

climbfuji:
But they use the ccpp type ccpp_t, which is defined in FV3/ccpp/framework/src/ccpp_types.F90 !

climbfuji:
```
use CCPP_data,         only: ccpp_suite,          &
                                cdata => cdata_tile, &
                                CCPP_interstitial
```
where cdata is from ccpp-framework (ccpp_t). Is this a direct or indirect dependency? Whatever, I'll remove the one line.

junwang-noaa:
I think that is implicit, fv3 dycore itself only needs ccpp_data and
ccpp_static_api

climbfuji:
> I think that is implicit, fv3 dycore itself only needs ccpp_data and ccpp_static_api
> […](#)

Agreed, I'll make this change and address @barlage's comments to the ccpp-physics PR when it these PRs are the next to commit.

climbfuji:
@dustinswales I added the rad_reset call for the gas concentrations that was missing in the last PR, and fixed a few RRTMGP issues in GFS_typedefs.meta. Please check commit 1d9c8487e2ebc765fd44df8fada49508fd733869. Thanks!

junwang-noaa (approved):

SMoorthi-emc (approved):
I assume the changes in radiation_surface are applicable only to NoahMP.

climbfuji:
> I assume the changes in radiation_surface are applicable only to NoahMP.

Yes, they are only used when the `ialbflg` parameter is set to 2. All regression tests that do not use Noah MP pass against the existing baselines with those changes.

PR #258, opened 2021-03-12
Fix model slowness when using threading

climbfuji:
## Description

This PR fixes the model slowness that we started to see after PR https://github.com/NOAA-EMC/fv3atm/pull/249 was merged. The reason are missing OpenMP dependencies in `ccpp/data/CMakeLists.txt` and `ccpp/driver/CMakeLists.txt`.

These dependencies were missing since the two folders `ccpp/data` and `ccpp/driver` were created. However, OpenMP was added implicitly through a dependency on target `gfsphysics`, which was removed in PR #249.

### Issue(s) addressed

ufs-weather-model issue [#474](https://github.com/ufs-community/ufs-weather-model/issues/474)
## Testing

Tested manually with Intel and GNU on Cheyenne and Hera for the regression test `fv3_ccpp_2threads` for 240h long integration times. With these changes the performance of the pre-#249 code is regained (and at the same time the memory footprint is reduced due to the changes in #249).

## Dependencies

n/a

### DIFF
--- a/ccpp/data/CMakeLists.txt
+++ b/ccpp/data/CMakeLists.txt
@@ -32,8 +32,12 @@ add_library(
     CCPP_data.F90
 )
 
-target_link_libraries(ccppdata ccpp)
-target_link_libraries(ccppdata ccppphys)
+target_link_libraries(ccppdata PUBLIC ccpp)
+target_link_libraries(ccppdata PUBLIC ccppphys)
+
+if(OpenMP_Fortran_FOUND)
+  target_link_libraries(ccppdata PUBLIC OpenMP::OpenMP_Fortran)
+endif()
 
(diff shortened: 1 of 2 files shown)


DusanJovic-NOAA (approved):

junwang-noaa (approved):

aerorahul:
were we not just still having a discussion on the `PUBLIC` and `PRIVATE` nature of OpenMP linking in another thread?

climbfuji:
Yes, but this can be done in my next commit (tomorrow) or later if we can't come to a conclusion by then. I will follow your advise as to what we should use, I don't understand cmake as well as you do.

junwang-noaa:
I think the slowness issue is resolved by the code changes. The update on
the PUBLIC and PRIVATE nature of OpenMP may require more changes besides
the changes in this PR, e.g. FV3/CMakeLists.txt,

PR #269, opened 2021-04-02
Add new PBL diagnostic variables dkt and dku

BinLiu-NOAA:
Add new PBL diagnostic variables dkt and dku (atmosphere heat and momentum diffusivity)

Contributors: @ChunxiZhang-NOAA and @AndrewHazelton, @BinLiu-NOAA 

This PR addresses [ccpp-physics issue #606 ](https://github.com/NCAR/ccpp-physics/issues/606).
Related PRs:
NCAR/ccpp-physics/pull/607
ufs-community/ufs-weather-model/pull/505

### DIFF
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1564,8 +1564,8 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: refl_10cm(:,:) => null()  !< instantaneous refl_10cm
 
     !--- Extra PBL diagnostics
-    real (kind=kind_phys), pointer :: dkt(:,:)       => null()  !< Eddy diffusitivity for heat from the EDMF and EDMF-TKE
-    real (kind=kind_phys), pointer :: dku(:,:)       => null()  !< Eddy diffusitivity for momentum from the EDMF and EDMF-TKE
+    real (kind=kind_phys), pointer :: dkt(:,:)       => null()  !< Eddy diffusitivity for heat
+    real (kind=kind_phys), pointer :: dku(:,:)       => null()  !< Eddy diffusitivity for momentum
 
 !
 !---vay-2018 UGWP-diagnostics instantaneous
@@ -1760,7 +1760,6 @@ module GFS_typedefs
     real (kind=kind_phys), pointer      :: del(:,:)           => null()  !<
     real (kind=kind_phys), pointer      :: del_gz(:,:)        => null()  !<
     real (kind=kind_phys), pointer      :: delr(:,:)          => null()  !<
-    real (kind=kind_phys), pointer      :: dkt(:,:)           => null()  !<
     real (kind=kind_phys), pointer      :: dlength(:)         => null()  !<
     real (kind=kind_phys), pointer      :: dqdt(:,:,:)        => null()  !<
     real (kind=kind_phys), pointer      :: dqsdt2(:)          => null()  !<
@@ -6294,7 +6293,6 @@ module GFS_typedefs
     allocate (Interstitial%del             (IM,Model%levs))
     allocate (Interstitial%del_gz          (IM,Model%levs+1))
     allocate (Interstitial%delr            (IM,Model%levr+LTP))
-    allocate (Interstitial%dkt             (IM,Model%levs-1))
     allocate (Interstitial%dlength         (IM))
     allocate (Interstitial%dqdt            (IM,Model%levs,Model%ntrac))
     allocate (Interstitial%dqsfc1          (IM))
@@ -6984,7 +6982,6 @@ module GFS_typedefs
     Interstitial%dd_mf           = clear_val
     Interstitial%del             = clear_val
     Interstitial%del_gz          = clear_val
-    Interstitial%dkt             = clear_val
     Interstitial%dlength         = clear_val
     Interstitial%dqdt            = clear_val
     Interstitial%dqsfc1          = clear_val
@@ -7334,7 +7331,6 @@ module GFS_typedefs
     write (0,*) 'sum(Interstitial%del             ) = ', sum(Interstitial%del             )
     write (0,*) 'sum(Interstitial%del_gz          ) = ', sum(Interstitial%del_gz          )
     write (0,*) 'sum(Interstitial%delr            ) = ', sum(Interstitial%delr            )
-    write (0,*) 'sum(Interstitial%dkt             ) = ', sum(Interstitial%dkt             )
     write (0,*) 'sum(Interstitial%dlength         ) = ', sum(Interstitial%dlength         )
     write (0,*) 'sum(Interstitial%dqdt            ) = ', sum(Interstitial%dqdt            )
     write (0,*) 'sum(Interstitial%dqsfc1          ) = ', sum(Interstitial%dqsfc1          )
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1564,7 +1564,8 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: refl_10cm(:,:) => null()  !< instantaneous refl_10cm
 
     !--- Extra PBL diagnostics
-    real (kind=kind_phys), pointer :: dkudiagnostic(:,:) => null()  !< Eddy diffusitivity from the EDMF and EDMF-TKE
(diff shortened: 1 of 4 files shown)


ccpp/data/GFS_typedefs.F90
```diff
@@ -5800,6 +5804,10 @@ subroutine diag_create (Diag, IM, Model)
     !--- 3D diagnostics for Thompson MP / GFDL MP
     allocate (Diag%refl_10cm(IM,Model%levs))
 
+    !--- New PBL Diagnostics
+    allocate (Diag%dkt(IM,Model%levs))
```
climbfuji: Shouldn't these be allocated to `Model%levs-1` to match the previous allocation in the interstitial DDT? If not, i.e. if you want to use `Model%levs` (for example to avoid issues when writing out the diagnostics), then you need to change the `dimensions` metadata in fv3atm and ccpp-physics, and the array declarations (the lines with `intent(in)`) from `dku(im,km-1)` to `dku(im,km)` (same for `dkt`) in both moninemf and satmedmfvdifq (and satmedmfvdif or any other scheme that is using either of these variables) .

climbfuji (approved):

ccpp/data/GFS_typedefs.F90
BinLiu-NOAA: Thanks, @climbfuji! Good catch. Indeed, it should be Model%levs-1. Have just changed accordingly.

In the meantime, I do see some of the atmosphere_heat/momentum_diffusivity* variables use Model%levs, while others use Model%levs-1. Probably also something keep in mind for future code standardization/unification.

Thanks!

climbfuji: Yes, please create an issue so that it doesn't get forgotten (unless you can remember, I certainly cannot). Does the diagnostic output still work without crashing, also in debug mode, when making this change?

BinLiu-NOAA: @climbfuji, Sure, I will create an issue either in FV3atm or ccpp/physics (please let me where you would prefer to open this issue, Dom) for potentially unifying the usage of different flavors (names, dimensions, etc.) of atmosphere_heat/momentum_diffusivity variables in FV3atm and ccpp/physics levels. Maybe this should apply to other variables as well. But, this will need some infrastructure level design/standard updates/changes.

For the latest feature/dkudiagnostic branch, @ChunxiZhang-NOAA is doing model-level regression tests now on Hera. Let's see how it goes with the latest changes. And Chunxi, please make sure you update your local branch with these latest changes. Thanks!

climbfuji: @BinLiu-NOAA thinking about this more, since there is no regression test for this, you will need to test manually: compile in debug mode, use an existing regression test that uses satmedmfvdifq or moninedmf, add the two arrays dku and dkt to `diag_table`, check if the code runs w/o out-of-bounds errors. Not clear how this will look like in the diagnostic files if these two arrays have one level less in the vertical dimension.

BrianCurtis-NOAA:
Please bring this branch up-to-date with NOAA-EMC:develop

ccpp/data/GFS_typedefs.F90
BinLiu-NOAA: @climbfuji, I have just brought this branch up to date with the latest develop branch. Let's first test to make sure the existing model-level RTs can go through. For the manual test you mentioned, could @ChunxiZhang-NOAA or @AndrewHazelton help to conduct that test. Not sure if we will be able to finish the above manual test you mentioned today though. Thanks!

AndrewHazelton: I am planning to test the update to fix the convective scheme issue in the global-nested version soon, so I can test dku and dkt then as well if you want.

ChunxiZhang-NOAA: I will do a regression test to see if these tow diag variables can be outputted correctly.  Then do a full regression test on Hera.

BinLiu-NOAA: Thanks @ChunxiZhang-NOAA! For the full model-level regression tests, I believe @BrianCurtis-NOAA is currently conducting that on Hera, I am doing it on Orion. So, you probably only need to conduct this specific test you and @climbfuji just mentioned to output the dku/dkt, maybe also test it under the debug mode. Thanks!

ChunxiZhang-NOAA: @BinLiu-NOAA  Sure. I just started working on this specific test on Hera.

ChunxiZhang-NOAA: I added dkt and dku in the diag table, the outputs looks fine except there are mixed zero and missing values at the top model level 64.  The run for the debug mode failed. The error message is: forrtl: severe (408): fort: (2): Subscript #2 of the array VAR3 has value 64 which is greater than the upper bound of 63.   We can ask Jongil if he would like to change the vertical dimension of dkt and dku from km-1 to km. I think he moved one level down to make all km-1 levels have values because the first level is zero.

climbfuji: That's what I expected to happen, thanks for confirming this @ChunxiZhang-NOAA. I think it is a good idea to extend `dku` and `dkt` to `km` in the vertical. It gets initialized to zero and stays zero for that level all the time.

BinLiu-NOAA: @ChunxiZhang-NOAA Thanks for testing this out! And @climbfuji, it looks to me that we should define and allocate dkt/dku at the FV3atm level as km, while leave them as km-1 in the ccpp/physics level. In this case, at the FV3atm level it will be properly defined, allocated, and initialized for all km levels. In the ccpp/physics level, only those km-1 levels will be used/updated. I suspect this probably will work fine, even with the debug mode. What do you think?

ccpp/data/GFS_typedefs.meta
```diff
@@ -6938,6 +6938,20 @@
   dimensions = (horizontal_loop_extent,vertical_dimension)
   type = real
   kind = kind_phys
+[dkt]
+  standard_name = atmosphere_heat_diffusivity
+  long_name = diffusivity for heat
```
junwang-noaa, line 6943: Maybe it is good to have the long_name be consistent with ExtDiag(idx)%desc, but we can do this later.

junwang-noaa (approved):
I hope the dkt/dku long name can be consistent with the output long name in ExtDiag(idx)%desc, but it can be done later.

ccpp/data/GFS_typedefs.F90
BinLiu-NOAA: Hi @climbfuji, @junwang-noaa, @BrianCurtis-NOAA, Following the above discussions, I made the following changes at the FV3atm level:

- Define and allocate dkt and dku (atmosphere heat/momentum diffusivity in GFS_typedefs.F90/meta with km levels.
- Unify the long name and ExtDiag%desc for dkt and dku (to address Jun's another comment).

More details can be seen from [this link](https://github.com/hafs-community/fv3atm/compare/feature/dkudiagnostic...hafs-community:feature/dktdku?expand=1). 
@ChunxiZhang-NOAA conducted a test with his previous failed debug test when outputting dkt and dku as diag variables, which can now go through successfully.

If you all agree, I can commit these changes into this Pull Request. Please let us know how you would like to proceed. Thanks!

junwang-noaa:
Bin, since all the RTs are is done, I suggest not to make further changes
this time, we can make the name change in next FV3 PR.

BinLiu-NOAA:
> Bin, since all the RTs are is done, I suggest not to make further changes this time, we can make the name change in next FV3 PR.

@junwang-noaa, This sounds good and reasonable to me. Thanks!

BrianCurtis-NOAA:
ccpp/physics has merged the changes. Please revert .gitmodules and update the pointer to NCAR:master for ccpp/physics.

climbfuji (approved):
ccpp-physics hash is correct. I created an issue to change the vertical dimension for the two arrays in ccpp-physics: https://github.com/NCAR/ccpp-physics/issues/639

junwang-noaa:
Bin, do we need to wait for Chunxi and Andrew to review the code or are we
ready to commit?

BinLiu-NOAA:
@junwang-noaa No, we do not need to wait for them. They are contributors for these changes. So, the PR is ready to go in this sense.

PR #272, opened 2021-04-07
fix lam post uninitialied fields

junwang-noaa:
## Description

This is to fix the uninitialized fields in lam post products, and the assignment of integer fields. The uninitialized fields cause run-ro-run reproducibility issue in LAM post test. 

### Issue(s) addressed

- fixes issue [ufs-weather-model #504](https://github.com/ufs-community/ufs-weather-model/issues/504) 
- fixes issue [ufs-weather-model #519](https://github.com/ufs-community/ufs-weather-model/issues/519) 

## Testing

The code has been tested on hera and orion. The run-to-run reproducibility issue is fixed. A final solution will be initializing all the allocated fields and  removing those fields in LAM post control file, which will be delivered from post group.

## Dependencies

-  ufs-weather-model PR

### DIFF
--- a/io/post_regional.F90
+++ b/io/post_regional.F90
@@ -448,7 +448,7 @@ module post_regional
       use vrbls3d,     only: t, q, uh, vh, wh, alpint, dpres, zint, zmid, o3,  &
                              qqr, qqs, cwm, qqi, qqw, qqg, omga, cfr, pmid,    &
                              q2, rlwtt, rswtt, tcucn, tcucns, train, el_pbl,   &
-                             pint, exch_h, ref_10cm
+                             pint, exch_h, ref_10cm, extcof55, aextc55, u, v
       use vrbls2d,     only: f, pd, sigt4, fis, pblh, ustar, z0, ths, qs, twbs,&
                              qwbs, avgcprate, cprate, avgprec, prec, lspa, sno,&
                              cldefi, th10, q10, tshltr, pshltr, tshltr, albase,&
@@ -477,7 +477,8 @@ module post_regional
                              up_heli_max03,up_heli_min03,rel_vort_max01,       &
                              rel_vort_max, rel_vort_maxhy1, refd_max,          &
                              refdm10c_max, u10max, v10max, wspd10max, sfcuxi,  &
-                             sfcvxi, t10m, t10avg, psfcavg, akhsavg, akmsavg
+                             sfcvxi, t10m, t10avg, psfcavg, akhsavg, akmsavg,  &
+                             albedo, tg
       use soil,        only: sldpth, sh2o, smc, stc
       use masks,       only: lmv, lmh, htm, vtm, gdlat, gdlon, dx, dy, hbm2, sm, sice
       use ctlblk_mod,  only: im, jm, lm, lp1, jsta, jend, jsta_2l, jend_2u, jsta_m,jend_m, &
@@ -641,7 +642,7 @@ module post_regional
           lspa(i,j) = SPVAL
           th10(i,j) = SPVAL
           q10(i,j) = SPVAL
-          albase(i,j) = SPVAL
+          albase(i,j) = 0.
         enddo
       enddo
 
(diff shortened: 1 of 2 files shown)


WenMeng-NOAA (approved):
These temp fix for initializing fields like t10m looks good to me.

io/post_regional.F90
```diff
@@ -829,6 +829,19 @@ subroutine set_postvars_regional(wrt_int_state,mpicomp,setvar_atmfile,   &
         enddo
       enddo
 !
+!** temporary fix: initialize t10m, t10avg, psfcavg, akhsavg, akmsavg
+!$omp parallel do default(none),private(i,j),shared(jsta_2l,jend_2u,im,spval), &
```
DusanJovic-NOAA: `spval` is not used in the following do loop.

io/post_regional.F90
```diff
@@ -1940,7 +1953,7 @@ subroutine set_postvars_regional(wrt_int_state,mpicomp,setvar_atmfile,   &
                 do i=ista, iend
                   if (arrayr42d(i,j) < spval) then
                     isltyp(i,j) = nint(arrayr42d(i,j))
-                    if( abs(arrayr42d(i,j)-fillValue) < small)  isltyp(i,j) = spval
+                    if( abs(arrayr42d(i,j)-fillValue) < small)  isltyp(i,j) = 0
```
DusanJovic-NOAA, line 1969: I assume 0 is not valid soil type, i.e. it's a 'special' value. Is that correct?

io/post_regional.F90
WenMeng-NOAA: I assume assigning 0 or spval have the same results in these fields. I have been working initializing all arrays in UPP code allocate_all.f. When that is done, this fix will be removed.

io/post_regional.F90
junwang-noaa: Right, the code follows standalone post to use zero for undefined value.

io/post_regional.F90
junwang-noaa: fixed

junwang-noaa:
Wen is going to have upp 10.0.6 release to fix the issues in this PR. Once upp 10.0.6 is installed on all the platforms, it will be used in this PR.

io/post_regional.F90
```diff
@@ -641,7 +642,7 @@ subroutine set_postvars_regional(wrt_int_state,mpicomp,setvar_atmfile,   &
           lspa(i,j) = SPVAL
           th10(i,j) = SPVAL
           q10(i,j) = SPVAL
-          albase(i,j) = SPVAL
+          albase(i,j) = 0.
```
WenMeng-NOAA, line 645: @junwang-noaa May I ask why albase is set as 0. instead of spval?

junwang-noaa:
It looks to me LAM does not output this field in history file. But with
"albase=spval", sometimes I got value 0. for this field and sometimes got
different values so I made a change in inline interface to avoid the issue,
please see below. I think array grid1 in line 237 in post code FIX.F90 is
not initialized, it may need to change to:

            IF(ABS(ALBASE(I,J)-SPVAL)>SMALL)  then
               GRID1(I,J) = ALBASE(I,J)*100.
            else
               GRID1(I,J) =spval
            endif


< 1078:37802593:vt=2018101500:surface:anl:SNFALB Snow-Free Albedo [%]:
<     ndata=53261:undef=674:mean=1.70765e+15:min=0:max=7.31e+19
---
> 1078:37762182:vt=2018101500:surface:anl:SNFALB Snow-Free Albedo [%]:
>     ndata=53261:undef=0:mean=0:min=0:max=0

WenMeng-NOAA:
@junwang-noaa Got it. I will add the fix in my upcoming UPP PR which would be applied in the next upp upgrade (10.0.7 or above).

RatkoVasic-NOAA (approved):

climbfuji (approved):

DusanJovic-NOAA (approved):

PR #301, opened 2021-05-19
Add import/export fields for fv3-jedi coupling

DusanJovic-NOAA:
## Description

Provide a detailed description of what this PR does.  
What bug does it fix, or what feature does it add?  
This PR adds import and export fields to support fv3-jedi coupling
Is a change of answers expected from this PR?  
No.

## Testing

How were these changes tested? Full regression tests on Hera and WCOSS Dell.
What compilers / HPCs was it tested with? Intel
Are the changes covered by regression tests? Not every code change is covered by ufs-weather-model tests, we do not run jedi in any of regression tests.
Have the ufs-weather-model regression test been run? On what platform?  Yes. Hera and Dell.
- Will the code updates change regression test baseline? No.
- Please commit the regression test log files in your ufs-weather-model branch.

## Dependencies

Depends on ufs-community/ufs-weather-model/pull/586

Do PRs in upstream repositories need to be merged first? No.

### DIFF
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -1518,11 +1518,13 @@ end subroutine atmos_data_type_chksum
     integer, intent(out) :: rc
 
     !--- local variables
-    integer :: n, j, i, ix, nb, isc, iec, jsc, jec, dimCount, findex
+    integer :: n, j, i, k, ix, nb, isc, iec, jsc, jec, nk, dimCount, findex
+    integer :: sphum, liq_wat, ice_wat, o3mr
     character(len=128) :: impfield_name, fldname
     type(ESMF_TypeKind_Flag)                           :: datatype
     real(kind=ESMF_KIND_R4),  dimension(:,:), pointer  :: datar42d
     real(kind=ESMF_KIND_R8),  dimension(:,:), pointer  :: datar82d
+    real(kind=ESMF_KIND_R8),  dimension(:,:,:), pointer:: datar83d
     real(kind=GFS_kind_phys), dimension(:,:), pointer  :: datar8
     real(kind=GFS_kind_phys)                           :: tem, ofrac
     logical found, isFieldCreated, lcpl_fice
@@ -1530,12 +1532,14 @@ end subroutine atmos_data_type_chksum
 !
 !------------------------------------------------------------------------------
 !
-! set up local dimension
     rc  = -999
+
+! set up local dimension
     isc = GFS_control%isc
     iec = GFS_control%isc+GFS_control%nx-1
     jsc = GFS_control%jsc
     jec = GFS_control%jsc+GFS_control%ny-1
+    nk  = Atm_block%npz
     lcpl_fice = .false.
 
     allocate(datar8(isc:iec,jsc:jec))
@@ -1548,7 +1552,6 @@ end subroutine atmos_data_type_chksum
     do n=1,nImportFields ! Each import field is only available if it was connected in the import state.
 
       found = .false.
-
 
       isFieldCreated = ESMF_FieldIsCreated(importFields(n), rc=rc)
       if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
@@ -1573,9 +1576,17 @@ end subroutine atmos_data_type_chksum
 !            call ESMF_FieldGet(importFields(n),farrayPtr=datar42d,localDE=0, rc=rc)
 !            datar8 = datar42d
           endif
+
+        else if( dimCount == 3) then
+          if ( datatype == ESMF_TYPEKIND_R8) then
+            call ESMF_FieldGet(importFields(n),farrayPtr=datar83d,localDE=0, rc=rc)
+            if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
+            found = .true.
+          endif
         endif
 !
-        if (found .and. datar8(isc,jsc) > -99998.0) then
+        if (found) then
+         if (datar8(isc,jsc) > -99998.0) then
 !
         ! get sea land mask: in order to update the coupling fields over the ocean/ice
 !        fldname = 'land_mask'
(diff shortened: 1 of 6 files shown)


junwang-noaa:
@rsdunlapiv you mentioned about the ESMF_METHOD_INITIALIZE, would you please show us what the changes need to make in the SetServices in fv3_cap? 
@rmontuoro Now we have two places to set up the coupled fields in fv3, assign_importdata/setup_exportdata and update_atmos_chemistry, I think the changes Dusan made for fve-judi should have no impact on fv3-chemistry coupling. Would you please double check? With the change, in your fv3-chem run, the shared 3D fields in fv3 export will be reset twice (in fv3 update_atmos_chemistry and then setup_exportdata), but should have no impact on fv3-chem run. The assign_importdata is only called by the namelist variable cplflx/cplwav, so it should have no impact either.

DusanJovic-NOAA:
@junwang-noaa thanks for noticing that `assign_importdata` is only called if one of the two cpl* flags is true. We need to change that to call it unconditionally, like we do with `setup_exportdata`.

junwang-noaa:
@DusanJovic-NOAA I am not sure if it will work with fv3-chem if you remove that. @rmontuoro can confirm. I think it will break the fv3-chem with the current run sequence as those fv3 tracer fields (state t=n) should not be updated from the fv3 import state (state t=n-1 after chem) in the place where assign_importdata is called. More code changes are required in order to remove the cplflx/cplwav/cplchem.

DusanJovic-NOAA:
we can then do this:
```
if( .not. GFS_control%cplchm ) then
  call assign_importdata(rc)
endif

```

junwang-noaa:
that should work.

atmos_model.F90
```diff
@@ -840,8 +826,8 @@ subroutine update_atmos_model_state (Atmos)
     call atmosphere_get_bottom_layer (Atm_block, DYCORE_Data)
 
     !if in coupled mode, set up coupled fields
```
climbfuji, line 828: This means for all runs, including `APP=ATM` atmosphere only runs, just not for coupled chemistry runs? Should the comments in lines 828 and 268 be updated?

climbfuji (approved):

rsdunlapiv:
Per @junwang-noaa question about, this is a good time to clean up the SetServices to use ONLY the semantic labels for each of the phases.  This is more clear than the "IPDv01p1" labels, which are hard to understand.

This means:
You don't need to register phase 0:
```
call ESMF_GridCompSetEntryPoint(gcomp, ESMF_METHOD_INITIALIZE,   userRoutine=InitializeP0, phase=0, rc=rc)
```

You don't need the subroutine `InitializeP0(gcomp, importState, exportState, clock, rc)` at all, except you would need to move the logic that is getting profile_memory, cplprint_flag, and dbug elsewhere.  (Make sure we actually need those anyway).

And the registration labels in `SetServices` can be changed:

```
call NUOPC_CompSetEntryPoint(gcomp, ESMF_METHOD_INITIALIZE, phaseLabelList=(/"IPDv01p1"/), userRoutine=InitializeAdvertise, rc=rc)
```
should be:
```
call NUOPC_CompSpecialize(model, specLabel=label_Advertise,  specRoutine=InitializeAdvertise, rc=rc)
```

And
```
call NUOPC_CompSetEntryPoint(gcomp, ESMF_METHOD_INITIALIZE, phaseLabelList=(/"IPDv01p3"/), userRoutine=InitializeRealize, rc=rc)
```
should be:
```
call NUOPC_CompSpecialize(model, specLabel=label_RealizeProvided,  specRoutine=InitializeRealize, rc=rc)
```

This also means that the signature of InitializeAdvertise and IntializeRealize can be simplified from:
```
subroutine InitializeAdvertise(gcomp, importState, exportState, clock, rc)
```
to
```
subroutine InitializeAdvertise(model, rc)    
    type(ESMF_GridComp)  :: model    
    integer, intent(out) :: rc
```

It is not critical that you do this now or in this PR, but it is a good cleanup and removes the confusing IPDxxxx stuff.

Another part of this - which should probably be a follow on, is to rework `InitializeAdvertise` because it is doing a ton of extra stuff that is not at all related to Advertise.  This is all the fcstgridcomp and io component logic.  I think it is a bigger project to take this on, and it would be related to a rework of the IO.

rsdunlapiv:
The full list of NUOPC semantic labels is here:
http://earthsystemmodeling.org/docs/nightly/develop/NUOPC_refdoc/node4.html#NUOPC_Model

atmos_model.F90
```diff
@@ -1852,6 +1849,385 @@ subroutine assign_importdata(rc)
...
+              do i=isc,iec
+                nb = Atm_block%blkno(i,j)
+                ix = Atm_block%ixp(i,j)
+                GFS_data(nb)%Sfcprop%zorl(ix) = datar82d(i-isc+1,j-jsc+1)
+              enddo
+            enddo
+          endif
+        endif
+
+        fldname = 't2m'
```
junwang-noaa, line 2215: I know JEDI-FV3 interface is using these coupling field names shared between the two components, but it may cause confusion in the future. I'd suggest to use the names in fd_nems.yaml which are shared by all the ufs subcomponents, I think 't2m" has a standard name "inst_temp_height2m". This can be fixed in a later PR.

atmos_model.F90
```diff
@@ -1926,683 +2302,659 @@ subroutine assign_importdata(rc)
...
+         else if (dimCount == 3) then
+           call ESMF_FieldGet(exportFields(n),farrayPtr=datar83d,localDE=0, rc=localrc)
+           if (ESMF_LogFoundError(rcToCheck=localrc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__, rcToReturn=rc)) return
+         else
+            write(0,*)'not implemented dimCount ',dimCount, trim(fieldname)
+            call ESMF_Finalize(endflag=ESMF_END_ABORT)
+         endif
+      else if (datatype == ESMF_TYPEKIND_R4) then
+         if (dimCount == 2) then
+           call ESMF_FieldGet(exportFields(n),farrayPtr=datar42d,localDE=0, rc=localrc)
```
junwang-noaa, line 2359: I see only datar83d is used later when setting up the pointers, I am not sure if this will work for 32BIT=Y, for which we may need to set up datar43d. It may not be an issue for fv3-mom6-cice6-wav coupling as those coupled fields are from physics which is always running with real8. We can update this later.

junwang-noaa (approved):

Issue #309, opened 2021-05-24
Enable coupling with GOCART component


rmontuoro:
## Description
Introduce changes required to enable coupling with UFS-Aerosols, a NUOPC-compliant component based on NASA GOCART.

junwang-noaa:
Code is committed, close issue.

PR #308, opened 2021-05-24
Fix segfault errors in coupled FV3-GOCART after PR #301

rmontuoro:
## Description

This PR refactors `setup_exportdata()` introducing Fortran preprocessor macros to minimize code replication.
The changes included prevent segfault errors observed when running the coupled FV3-GOCART system (#307).

### Issue(s) addressed
- fixes #307 

## Testing

Successful regression tests performed on:
- hera.intel
- orion.intel
- wcoss_dell_p3

Regression test logs available at: https://github.com/rmontuoro/ufs-weather-model/tree/feature/gocart

This PR does not introduce baseline changes.

## Dependencies
None.

PR #310, opened 2021-05-24
Enable GOCART integration in UFS

### LINKED ISSUE #307: Coupled FV3-GOCART system crashes after PR #301
## Description
PR #301 crashes FV3 when coupling with GOCART due to updates in `setup_exportdata()`, which reference unallocated pointers when `cplflx` is .false.

### To Reproduce:
Issue observed on Hera and Orion with the Intel compiler when running with the coupled GOCART component (`cplchem = .true.`, `cplflx = .false.`)

rmontuoro:
## Description

This PR adds the required infrastructure to couple with UFS-Aerosols, a prognostic aerosol model based on NASA GOCART (see #309 and UFS weather model issue [#590](https://github.com/ufs-community/ufs-weather-model/issues/590)).

It also includes a hotfix to prevent segfault errors when coupling with chemistry (see PR #308)

Features:
- Implemented sharing of atmospheric tracer metadata (name, units) to coupled model component via a standard NUOPC connector if coupling tracer field is accessed by shared memory reference
- Introduced additional metadata to tracer records in field_table to identify and validate prognostic and diagnostic chemistry tracers. A new record line starting with the `tracer_usage` keyword is now required to recognize chemistry tracer as follows:
```
      "tracer_usage", "chemistry"                    (default, prognostic)
      "tracer_usage", "chemistry", "type=diagnostic" (diagnostic)
```
- Added sanity check for input chemistry tracers, ensuring that both prognostic and diagnostic tracers in field_table are contiguous, and diagnostic tracers follow prognostic ones.
- Improved algorithm to set aerosol scavenging factors
- Exported instantaneous fields for:
   - 3D non-convective liquid and ice precipitation fluxes
   - 3D cloud fraction
   - normalized soil wetness
   -  ice fraction (as used in the atmospheric model)
   - lake fraction
   - ocean fraction
   - surface snow area fraction
- Enabled import of diagnostic tracers

Fixes:
- Prevent a divide by zero error when costant arrays are rescaled for lossy compression in `write_netcdf_parallel()`
- Correctly sample and compute elapsed time for field bundle regridding operations.
- Include hotfix (see PR #308) to allow latest `setup_exportdata()` to work with coupled chemistry.

This PR also eliminates legacy code used for GEFS-Aerosols coupling and no longer needed.

### Issue(s) addressed

- addresses #309 
- fixes #307 

## Testing

Successful regression tests performed for:
- hera.intel
- orion.intel
- wcoss_dell_p3

Log files available at: https://github.com/rmontuoro/ufs-weather-model/tree/feature/gocart

This PR does not change the regression test baseline

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/664


atmos_model.F90
```diff
@@ -929,49 +935,113 @@ subroutine get_atmos_model_ungridded_dim(nlev, nsoillev, ntracers,     &
...
 
-  !--- number of tracers used in emission diagnostic output
-  if (present(num_diag_burn_emis_flux)) then
-    num_diag_burn_emis_flux = 0
-    if (associated(GFS_data(1)%IntDiag%abem)) &
-      num_diag_burn_emis_flux = size(GFS_data(1)%IntDiag%abem, dim=2)
+  if (id_num > 0) then
+    !--- check if diagnostic tracers are contiguous
+    if (id_num > id_max - id_min + 1) &
+      call mpp_error(FATAL, 'diagnostic chemistry tracers must be consecutive')
```
climbfuji: Change to `contiguous` for consistency with line 1031?

ccpp/data/GFS_typedefs.F90
```diff
@@ -4277,59 +4269,12 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
-            if (n > 0) Model%fscav(n) = tem
-          endif
-        endif
-      enddo
-    endif
+!--- initialize parameters for atmospheric chemistry tracers
+    call Model%init_chemistry(tracer_types)
+
+!--- setup aerosol scavenging factors
+    call Model%init_scavenging(fscav_aero)
```
climbfuji, line 4341: Good to see this being cleaned up.

climbfuji (approved):
I cannot comment on the `ESMF` coupling side of things, the rest looks good to me.

include/atm_data_copy.h
```diff
@@ -0,0 +1,64 @@
...
+#  define __beg_k_loop
+#  define __end_k_loop
+#endif
+
+#undef __lev
+#ifdef _atm_copy_lev
+#  define __lev ,:
+#else
+#  define __lev
+#endif
```
DusanJovic-NOAA: I do not like all this ifdef/undef/define magic. This is very unreadable, difficult to debug and maintain. We should not use this kind of cpp string processing to try to write generic code. This should all be written in Fortran.

atmos_model.F90
rmontuoro: Done in commit c48922a

include/atm_data_copy.h
rmontuoro: @DusanJovic-NOAA I generally agree with you, but there are cases where writing the same (or almost the same) code fragment a large number of times in a single file increases the chances of introducing a bug. It also makes debugging and maintenance quite difficult since it would involve visually parsing almost identical code blocks.

The issue here is that we cannot define a Fortran function to replace each code block and improve modularity as well as readability. This case includes 74 almost identical code fragments and more may be added in the future. Using smaller building blocks that can be quickly updated and tested will improve resiliency for the entire code base.

DusanJovic-NOAA: All you need is couple of fortran subroutines that accept 2d, 3d or 4d pointers, have a 2d, 3d or 4d do-loops and copy the values to the destination. I agree that we do not want to repeat all those 74 loops 74 times. But like I said using all those ifdefs/undefs/defines to achieve some kind of generality is not the way to go. I find this code extremely unreadable and difficult to reason about and debug.

rmontuoro: @DusanJovic-NOAA Could you please provide an example of such subroutine? I'd welcome a simpler solution.

The issue I see is that we can't pass each source array as an argument to a generic subroutine, since these arrays are components of a derived type array (_e.g._ `GFS_Data`) and we need to loop over each of its elements (number of grid blocks) to retrieve them.

DusanJovic-NOAA: I believe atm_data_copy can be written as a simple subroutine, or few generic subroutines. gfs_data_copy, because of the GFS_data blocking, probably can not. But even so, I'd rather deal with (repeated) code like this:

```
!$omp parallel do default(shared) private(i,j,nb,ix)                                                                                 
      do j=jsc,jec                                                                                                                   
        do i=isc,iec                                                                                                                 
          nb = Atm_block%blkno(i,j)                                                                                                  
          ix = Atm_block%ixp(i,j)                                                                                                    
          datar82d(i-isc+1,j-jsc+1) = GFS_data(nb)%coupling%snow_cpl(ix) * rtimek                                                    
        enddo                                                                                                                        
      enddo                                                                                                                          
```
than something like:
```
          ! MEAN snow precipitation rate (kg/m2/s)
          case ('mean_fprec_rate')
#define _atm_src_ptr coupling%snow_cpl
#include "gfs_data_copy.h"
```
and hope that everyone understands that several lines above, somebody correctly defines:
```
#undef  _atm_src_cnv
#define _atm_src_cnv rtimek
```
so that scaling is done properly.
And what if somebody adds a new field after `mean_fprec_rate`, that does not have to be scaled, but forgets to `#undef __atm_src_cnv`. Constantly defining and un-defining all those cpp macros is in my opinion very error prone, much more than bunch of repeated loops, where at least we can see exactly what's going on. What if one needs more complicated scaling than simple multiplication factor. 

How would you debug (print) just one field. 

@junwang-noaa @climbfuji @DeniseWorthen What's your opinion on this?

junwang-noaa:
I agree. I was looking at those include files and find it's not easy to
follow up with these undef/define lines. Can we create a Fortran subroutine
for these calls?

include/atm_data_copy.h
rmontuoro: I agree that documentation needs to be added for the macros.

DeniseWorthen:
I agree w/ both Dusan and Jun here. I can't really follow what is happening w/ the various define/undefine statements.

rmontuoro:
I understand preprocessor macros may not be ideal. However, I would encourage everyone to find a solution preserving code modularity and following best practices, which would pay off in the long term.

Adding hundreds of lines of mostly identical code to a single file will make the code base difficult to maintain and debug. Using few individual procedures that can be independently validated as well as integrated in unit tests would increase robustness and help prevent bugs.

junwang-noaa:
I'd suggest creating subroutines to assign the fields according to their
data structure, instead of these include files, we can then call the fields
through subroutines. I am not sure if the include files will be a better
option for unit test than the subroutines.

rmontuoro:
One possible approach using Fortran subroutines could be the following:

```
module data_copy_mod
  ...
  interface gfs_data_copy
    module procedure gfs_data_copy2d
    module procedure gfs_data_copy3d
  end interface

contains

  subroutine gfs_data_copy2d(array, data, atm_block, nb)
    real, intent(out) :: array(:,:)
    real, intent(in)  :: data(:,:)
    type (block_control_type), intent(in) :: atm_block
    integer, intent(in) :: nb
    integer :: i, ib, ix, j, jb
    do ix = 1, Atm_block%blksz(nb)
      ib = Atm_block%index(nb)%ii(ix)
      jb = Atm_block%index(nb)%jj(ix)
      i = ib - Atm_block%isc + 1
      j = jb - Atm_block%jsc + 1
      array(i,j) = data(ix)
    enddo
  end subroutine gfs_data_copy2d

  subroutine gfs_data_copy3d(array, data, atm_block,nb)
    real, intent(out) :: array(:,:,:)
    real, intent(in)  :: data(:,:,:)
    type (block_control_type), intent(in) :: atm_block
    integer, intent(in) :: nb
    integer :: i, ib, ix, j, jb, k
    do k = 1, Atm_block%npz
      do ix = 1, Atm_block%blksz(nb)
        ib = Atm_block%index(nb)%ii(ix)
        jb = Atm_block%index(nb)%jj(ix)
        i = ib - Atm_block%isc + 1
        j = jb - Atm_block%jsc + 1
        array(i,j,k) = data(ix,k)
      enddo
    enddo
  end subroutine gfs_data_copy3d

end module
```

```
use data_copy_mod
...
!$omp parallel do default (shared) private (nb)
  do nb = 1, Atm_block%nblks
    select case (fieldName)
      case ('inst_zonal_wind_height10m')
        call gfs_data_copy(data82d, GFS_data(nb)%coupling%u10mi_cpl, Atm_block,nb)
      case ('inst_pres_levels')
        call gfs_data_copy(data83d, GFS_data(nb)%Statein%prsl, Atm_block,nb)
      ...
    end select
  enddo
```

junwang-noaa:
Yes, this is what I mean.

rmontuoro:
Thanks @junwang-noaa. @DusanJovic-NOAA and @DeniseWorthen, do you agree with this approach?

DeniseWorthen:
@rmontuoro This looks much more understandable to me. Thanks.

DusanJovic-NOAA:
> Thanks @junwang-noaa. @DusanJovic-NOAA and @DeniseWorthen, do you agree with this approach?

Yes. This is much better.

rmontuoro:
Commit 03f841d8bed04ea4370bf21a530f3ad6e0ea5019 (also included in PR #308) introduces generic Fortran subroutines to streamline `setup_exportdata()`, replacing previously implemented Fortran macros as agreed.

CMakeLists.txt
```diff
@@ -143,6 +144,7 @@ list(APPEND _fv3atm_defs_private GFS_PHYS
 target_compile_definitions(fv3atm PRIVATE "${_fv3atm_defs_private}")
 
 set_target_properties(fv3atm PROPERTIES Fortran_MODULE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/mod)
+target_include_directories(fv3atm PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include)
```
DusanJovic-NOAA: Not needed anymore.

atmos_model.F90
```diff
@@ -2337,932 +2353,252 @@ subroutine setup_exportdata()
...
+              call block_data_copy(datar83d, GFS_data(nb)%sfcprop%smc, Atm_block, nb)
+            case ('snwdph')
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%snowd, Atm_block, nb)
+            case ('f10m')
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%f10m, Atm_block, nb)
+            case ('zorl')
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%zorl, Atm_block, nb)
+            case ('t2m')
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%t2m, Atm_block, nb)
+          end select
```
DusanJovic-NOAA, line 2693: Should we have a default case with a warning message or an error?

rmontuoro: Since the coupling fields may or may not be available, there is no reference configuration we can use to issue warnings.
The generic subroutines check if both source and destination pointers are associated before attempting to copy the data to prevent cases where not all listed fields are connected (which would crash the model).

DusanJovic-NOAA: If the field is realized, that means FV3 advertised it and some other component is expecting 'valid' values. If we can not provide the values it is either logic error or a bug. In either case we should at least have some warning, if not an error.

rmontuoro:
@DusanJovic-NOAA - True, but these methods could also be used in a different context.

DusanJovic-NOAA:
> @DusanJovic-NOAA - True, but these methods could also be used in a different context.

In which context?

rmontuoro:
They could be used to build or update other procedure across FV3. For instance, some of the existing loops could be replaced with a call to `block_data_copy()` or `block_atmos_copy()` to streamline the FV3 code base. Each of these methods could also be incorporated into unit tests in the future.

CMakeLists.txt
rmontuoro: Commit 7d331b4f3883dac990eaae1d4f7a45768027ed1d removes the no longer needed line. Thanks @DusanJovic-NOAA!

DusanJovic-NOAA:
> They could be used to build or update other procedure across FV3. For instance, some of the existing loops could be replaced with a call to `block_data_copy()` or `block_atmos_copy()` to streamline the FV3 code base. Each of these methods could also be incorporated into unit tests in the future.

In all those cases both destination and source should be associated, if they are not then why would caller call this subroutine. But, if you want to allow caller to call it with a null pointer, then there should be a warning, or non-zero return code. How are we going to know that copy is not made, how do we catch bugs, logic errors, typos?

Having a routine that is supposed to copy data, but for whatever reason is not capable of making a copy, should be some kind of exception. It didn't do what it's supposed to do. Routine must not silently return.

Previously, with explicit do loops I could use compiler flags to help me find bugs, now with these if tests, code will silently continue if I inadvertently pass null pointer.

cpl/module_block_data.F90
```diff
@@ -0,0 +1,483 @@
...
+    real(ESMF_KIND_R8),        pointer    :: destin_ptr(:,:,:)
+    real,                      intent(in) :: source_arr(:,:,:,:)
+    integer,                   intent(in) :: slice
+    type (block_control_type), intent(in) :: block
+    integer,                   intent(in) :: block_index
+    real, optional,            intent(in) :: scale_factor
+
+    ! -- local variables
+    integer         :: i, ib, ix, j, jb, k
+    real(kind_phys) :: factor
```
DusanJovic-NOAA: Why is `factor`'s type `kind_phys`, when both `scale_factor` and `source_arr` are default real?

cpl/module_block_data.F90
```diff
@@ -0,0 +1,483 @@
...
+        jb = block%index(block_index)%jj(ix)
+        i = ib - block%isc + 1
+        j = jb - block%jsc + 1
+        destin_ptr(i,j) = factor * source_ptr(ib,jb)
+      enddo
+    end if
+
+  end subroutine block_copy_2d_to_2d_r8
+
+  subroutine block_array_copy_2d_to_2d_r8(destin_ptr, source_arr, block, block_index, scale_factor)
```
DusanJovic-NOAA: Does `_r8` in the subroutine name refers to source or destination kind?

rmontuoro: The destination kind. So far the source kind is always real*4.
We can certainly add the source kind to the procedure's name.

cpl/module_block_data.F90
rmontuoro: Good point! Will fix it.

cpl/module_block_data.F90
DusanJovic-NOAA: Is it? Looks to me source kind can be either 4 or 8 depending on how we build the model. If you want this routine to be called only for 4-byte reals then explicitly set the kind of source_arr.

cpl/module_block_data.F90
rmontuoro: Commit 2adafb484cfadfadf6fdcfc2cd4624174b713253 fixed this issue.

cpl/module_block_data.F90
rmontuoro: You are right, I should have been more clear. The parametric kind dependency takes care of the source argument, so we don't need to replicate more code.

rmontuoro:
> > They could be used to build or update other procedure across FV3. For instance, some of the existing loops could be replaced with a call to `block_data_copy()` or `block_atmos_copy()` to streamline the FV3 code base. Each of these methods could also be incorporated into unit tests in the future.
> 
> In all those cases both destination and source should be associated, if they are not then why would caller call this subroutine. But, if you want to allow caller to call it with a null pointer, then there should be a warning, or non-zero return code. How are we going to know that copy is not made, how do we catch bugs, logic errors, typos?
> 
> Having a routine that is supposed to copy data, but for whatever reason is not capable of making a copy, should be some kind of exception. It didn't do what it's supposed to do. Routine must not silently return.
> 
> Previously, with explicit do loops I could use compiler flags to help me find bugs, now with these if tests, code will silently continue if I inadvertently pass null pointer.

You may add logic checking that all pointers are associated outside the call if appropriate in your context.
Errors or warnings are meaningless in `setup_exportdata()` since the pointer status depends on the coupling configuration, which is determined at runtime.

DusanJovic-NOAA:
> > > They could be used to build or update other procedure across FV3. For instance, some of the existing loops could be replaced with a call to `block_data_copy()` or `block_atmos_copy()` to streamline the FV3 code base. Each of these methods could also be incorporated into unit tests in the future.
> > 
> > 
> > In all those cases both destination and source should be associated, if they are not then why would caller call this subroutine. But, if you want to allow caller to call it with a null pointer, then there should be a warning, or non-zero return code. How are we going to know that copy is not made, how do we catch bugs, logic errors, typos?
> > Having a routine that is supposed to copy data, but for whatever reason is not capable of making a copy, should be some kind of exception. It didn't do what it's supposed to do. Routine must not silently return.
> > Previously, with explicit do loops I could use compiler flags to help me find bugs, now with these if tests, code will silently continue if I inadvertently pass null pointer.
> 
> You may add logic checking that all pointers are associated outside the call if appropriate in your context.
> Errors or warnings are meaningless in `setup_exportdata()` since the pointer status depends on the coupling configuration, which is determined at runtime.

Passing a null pointer can be due to a logic error, or a bug or a typo. I want this routine to catch those situations.
I do not know in which coupling situations null pointer is a valid status of either source or destination, but then in those situations checking should be done outside to avoid warning if you do not like seeing the warning. I want to see a warning if I make a typo,
routine should not silently continue in those cases. For example what will happen if I type:
```
            case ('mean_prec_rate')
              call block_data_copy(datar42d, GFS_data(nb)%coupling%rain_cpl, Atm_block, nb, scale_factor=rtimek)
```
instead of:
```
            case ('mean_prec_rate')
              call block_data_copy(datar82d, GFS_data(nb)%coupling%rain_cpl, Atm_block, nb, scale_factor=rtimek)
```

rmontuoro:
It's hard to code against such situations, since it is the developer's responsibility to implement either of the two equally valid choices.

Again, pointers from fields in both import/export states, as well as in fv3atm internal's data structures are associated based on the coupling configuration (_e.g._ `cplflx=.true./.false.`, etc.), which is chosen at runtime. Issuing a warning due to an unallocated pointer resulting from a perfectly valid coupling configuration would be misleading. Issuing an error and stopping the model would be wrong.

The bug we are trying to fix originated exactly from using a pointer that was correctly unassociated in a valid coupling configuration that was not tested before.

DusanJovic-NOAA:
> It's hard to code against such situations, since it is the developer's responsibility to implement either of the two equally valid choices.
> 
> Again, pointers from fields in both import/export states, as well as in fv3atm internal's data structures are associated based on the coupling configuration (_e.g._ `cplflx=.true./.false.`, etc.), which is chosen at runtime. Issuing a warning due to an unallocated pointer resulting from a perfectly valid coupling configuration would be misleading. Issuing an error and stopping the model would be wrong.
> 
> The bug we are trying to fix originated exactly from using a pointer that was correctly unassociated in a valid coupling configuration that was not tested before.

We are talking here about fields in an export state.

I guess I do not understand "_a pointer that was correctly unassociated in a valid coupling configuration_".

Are you saying that in situation where export fields that FV3 advertised and that were realized, which means some other component relies on them being actually exported from FV3, are not being exported because some pointers are null due to some runtime logical flag. I would call this run-time miss-configuration, which we should catch with warnings or errors.

How would other component know (detect) such situation? What values will be passed in an export field to other component, a zero, some other special values?

rmontuoro:
Yes, that's what I'm saying, but this can also occur with valid coupling configurations.

Ideally, internal data pointers (_e..g._:` GFS_data(:)%Coupling%pointer(:)`) should be allocated and populated based on whether the corresponding exported field(s) are realized. However, at this time, these two entities behave independently: internal pointers are allocated (and populated) according to logical flags in the `&gfs_physics_nml` namelist (`cplflx`, `cplwav`, `cplchm`, etc), while coupling fields are realized depending on the coupled components. 

This situation is prone to failures and should be addressed. However, the complexity arising from the increasingly large number of coupling configuration makes finding a reasonable solution challenging.

You may try a quick experiment: comment out the lines protecting from unassociated pointers in `cpl/module_block_data.F90`:
```
-    if (associated(destin_ptr) .and. associated(source_ptr)) then
+!   if (associated(destin_ptr) .and. associated(source_ptr)) then
```
then run with all `cpl*` flags set to `.false.` except for `cplchm=.true.`. This is a valid coupling configuration (atmosphere-chemistry). The model will crash (SEGV) in `setup_exportdata()` because not all referenced internal pointers are associated.

We should work together to minimize these issues, but in the meantime it is crucial to guard against unassociated pointers when working with FV3.

DusanJovic-NOAA:
> Yes, that's what I'm saying, but this can also occur with valid coupling configurations.
> 
> Ideally, internal data pointers (_e..g._:` GFS_data(:)%Coupling%pointer(:)`) should be allocated and populated based on whether the corresponding exported field(s) are realized. However, at this time, these two entities behave independently: internal pointers are allocated (and populated) according to logical flags in the `&gfs_physics_nml` namelist (`cplflx`, `cplwav`, `cplchm`, etc), while coupling fields are realized depending on the coupled components.
> 
> This situation is prone to failures and should be addressed. However, the complexity arising from the increasingly large number of coupling configuration makes finding a reasonable solution challenging.
> 
> You may try a quick experiment: comment out the lines protecting from unassociated pointers in `cpl/module_block_data.F90`:
> 
> ```
> -    if (associated(destin_ptr) .and. associated(source_ptr)) then
> +!   if (associated(destin_ptr) .and. associated(source_ptr)) then
> ```
> 
> then run with all `cpl*` flags set to `.false.` except for `cplchm=.true.`. This is a valid coupling configuration (atmosphere-chemistry). The model will crash (SEGV) in `setup_exportdata()` because not all referenced internal pointers are associated.
> 
> We should work together to minimize these issues, but in the meantime it is crucial to guard against unassociated pointers when working with FV3.

This is exactly the situation I want to catch with hard errors, not just warnings or doing nothing. If field is realized it must be exported, that's what we promised. If internal pointer holding values for such field is not allocated that's a bug.  A bug which must be fixed. Not avoided by adding all these if tests that just silently skip over null pointer, and just let the code execute like nothing happened.

We must not used the excuse that "_complexity arising from the increasingly large number of coupling configuration_" and let potentially serious and hard to notice bugs go undetected because "_finding a reasonable solution is challenging_". This is bad practice that I want to avoid.

Remove all those if null pointer checks, let the model crash, and let's fix the bugs.

rmontuoro:
@DusanJovic-NOAA - I fully agree with you. I would point out that refactoring FV3 to ensure that every exported field has a corresponding associated internal pointer regardless of the status of the logical coupling flags (`cplflx`, etc.) is well outside the scope of this PR, as is the commit we are trying to address, which was introduced with PR #301. I had started a conversation last year within EMC on this general issue and would be happy to participate in the effort.

This PR highlighted two existing issues with `setup_exportdata()`:

1. The exported `inst_land_sea_mask` field points to `GFS_data%coupling%slmsk_cpl` which is never populated. It should instead point to `GFS_data%sfcprop%slmsk`, as done for the `slmsk` field. 
2. The `inst_sensi_heat_flx` field points to `GFS_data%coupling%dtsfci_cpl`, which is associated only if `cplflx=.true.` The coupling configuration for the atmosphere-chemistry model requires `cplflx=.false.` and uses the identical coupling field, causing a SIGV. Furthermore, the chemistry component requires the instantaneous sensible heat flux to be defined on the entire atmospheric grid, while `dtsfci_cpl` is only computed over the ocean. A different internal quantity (`GFS_Data%Coupling%ushfsfci`) was introduced early on with GSDCHEM, which was exported using the `inst_up_sensi_heat_flx` field.
We now have two options:
  (a) Continue using different field names for the exported instantaneous sensible heat flux defined on the entire grid (`inst_up_sensi_heat_flx`) and the same flux defined only over the ocean (`inst_sensi_heat_flx`)
  (b) Populate `inst_sensi_heat_flx` (`dtsfci_cpl`) on the full atmospheric grid

I'd be happy to implement either one of those solutions if you agree. @junwang-noaa and @DeniseWorthen, what's your preference?

DeniseWorthen:
We have just implemented the normalization for CMEPS to reflect the fact that 5 of the fluxes exported by ATM->OCN are calculated over the open water only. There are 5 of these fields (lwnet, sensible, latent and the u,v momentum fluxes). I believe it makes more sense to maintain these fields as is for the fully coupled ufs-weather. So my preference would be 2a.

rmontuoro:
> We have just implemented the normalization for CMEPS to reflect the fact that 5 of the fluxes exported by ATM->OCN are calculated over the open water only. There are 5 of these fields (lwnet, sensible, latent and the u,v momentum fluxes). I believe it makes more sense to maintain these fields as is for the fully coupled ufs-weather. So my preference would be 2a.

Thanks, @DeniseWorthen. This makes option 2a the best option at this time. I will provide an update based on this option.

rmontuoro:
The latest updates implement option 2a and enforce pointer association as agreed. @DusanJovic-NOAA, @DeniseWorthen , and @junwang-noaa please let me know if this works for you

atmos_model.F90
```diff
@@ -2390,217 +2390,220 @@ subroutine setup_exportdata(rc)
...
-              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%f10m, Atm_block, nb)
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%f10m, Atm_block, nb, rc=localrc)
             case ('zorl')
-              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%zorl, Atm_block, nb)
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%zorl, Atm_block, nb, rc=localrc)
             case ('t2m')
-              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%t2m, Atm_block, nb)
+              call block_data_copy(datar82d, GFS_data(nb)%sfcprop%t2m, Atm_block, nb, rc=localrc)
+            case default
+              localrc = ESMF_SUCCESS
```
DusanJovic-NOAA: The default case should be an error, not success.

rmontuoro: @DusanJovic-NOAA - `setup_exportdata()` is not comprehensive. For instance, exported fields used by the chemistry model are not included in this subroutine and should not be at this time since they are exported at the end of FV3 run phase 1. 
`setup_exportdata()` is used during FV3 initialization and by `update_atmos_model_state()`, which is called in run phase 2, except when FV3 is coupled with chemistry.

I understand the motivation, but setting the default case to be an error would break the coupled model.
Again, refactoring this code should not be part of this PR.

DusanJovic-NOAA: `setup_exportdata` is not used if `cplchem` is true, so the fact that not all chemistry fields are exported is irrelevant.

Why would it break the coupled model? Are there still exported fields (other than the 'chemistry' fields) that are not handled by the 'select case'?

rmontuoro: `setup_exportdata()` is called by `InitializeRealize()` (`fv3_cap.F90`) even if `cplchm` is `.true.`, and this is where all those issues are coming from.

If `setup_exportdata()` was not intended to be used with the coupled chemistry model, I'd be happy to disable it when `cplchm` is set to `.true.`. Please let me know.

DusanJovic-NOAA: I do not know if you need to call `setup_exportdata` with the chemistry model from `InitializeRealize` or not. But if you do then add all exported fields, and set the default case to error. If the default case is not an error there's no way to catch typos. For example, if instead:
`            case ('o3mr')`
I write:
`            case ('o3nr')`
model will silently continue to run, without me noticing that o3mr is not exported.

rmontuoro: `setup_exportdata()` is not required at all when coupling with the chemistry model. I do agree with you that, generally speaking, we should return an error when a mismatch occurs.

I suggest: 
1. checking if `cplchm` is `.true.` inside `setup_exportdata()` and bailing out if so (this is needed since `GFS_control%cplchm` is not accessible in the FV3 cap)
2. returning with an error in the default case, as you suggest.

Would this work?

DusanJovic-NOAA: I think it should.

rmontuoro: @DusanJovic-NOAA - I've implemented the changes above. Please take a look.

junwang-noaa:
I replied in email, but I was told there is a temporary issue with github. I reposted the message here.

Raffaele, I think we discussed this at the last fv3-jedi meeting, we should use one single interface setup_exportdata for all the applications including fv3-mom6/cice6/ww3, fv3-jedi and fv3-chem. If there are fields used by several applications with the same names but associated with different fv3 internal fields, we can use namelist variables (e.g. cplchm) to set up the correct field value. Also the setup_exportdata can be called at  InitializeRealize() if the internal coupled fields are properly initialized. It will be called again at the end of fv3 forecast phase1, so you will get updated fv3 export fields before forecast phase2 is called, I don't think there is any issue there. Using both setup_exportdata and update_atmos_chemistry  for setting up fv3 export fields are causing confusion. @rsdunlapiv @@theurich @mark-a-potts can comment on this.

climbfuji:
I updated the ccpp-physics submodule pointer following the merge of https://github.com/NCAR/ccpp-physics/pull/664 and reverted the change to `.gitmodules`. Please check @DusanJovic-NOAA @junwang-noaa @rmontuoro

DusanJovic-NOAA (approved):
Looks good.

atmos_model.F90
```diff
@@ -2431,932 +2447,254 @@ subroutine setup_exportdata()
...
+              call block_data_copy_or_fill(datar82d, DYCORE_data(nb)%coupling%v_bot, zeror8, Atm_block, nb, rc=localrc)
+            ! bottom layer pressure (p)
+            case('inst_pres_height_lowest')
+              call block_data_copy_or_fill(datar82d, DYCORE_data(nb)%coupling%p_bot, zeror8, Atm_block, nb, rc=localrc)
+            ! bottom layer height (z)
+            case('inst_height_lowest')
+              call block_data_copy_or_fill(datar82d, DYCORE_data(nb)%coupling%z_bot, zeror8, Atm_block, nb, rc=localrc)
+            !--- JEDI fields
+            case ('u')
+              call block_atmos_copy(datar83d, Atm(mygrid)%u, Atm_block, nb, rc=localrc)
```
junwang-noaa, line 2640: Do you run fv-gocart test with 64bit?  the Atm(mygrid)%u can be real(4) is compiled with 32BIT=Y.

io/module_write_netcdf_parallel.F90
```diff
@@ -378,15 +378,20 @@ subroutine write_netcdf_parallel(fieldbundle, wrtfb, filename, mpi_comm, mype, i
...
-              allocate(arrayr4_3d_save(i1:i2,j1:j2,k1:k2))
-              arrayr4_3d_save(i1:i2,j1:j2,k1:k2)=arrayr4_3d(i1:i2,j1:j2,k1:k2)
-              arrayr4_3d = scale_fact*(nint((arrayr4_3d_save - offset) / scale_fact)) + offset
-              ! compute max abs compression error.
-              compress_err(i) = &
-              maxval(abs(arrayr4_3d_save(i1:i2,j1:j2,k1:k2)-arrayr4_3d(i1:i2,j1:j2,k1:k2)))
-              deallocate(arrayr4_3d_save)
-              call mpi_allreduce(mpi_in_place,compress_err(i),1,mpi_real4,mpi_max,mpi_comm,ierr)
-              !print *,'field name=',trim(fldName),dataMin,dataMax,compress_err(i)
+              if (scale_fact > 0.) then
```
junwang-noaa, line 381: Is original code causing issues or the code changes are to skip the allocating array and mpi calls for constant field?

junwang-noaa (approved):
The code changes are working. Please see some questions of mine.

io/module_write_netcdf_parallel.F90
rmontuoro: This change is meant to address an issue in the original code.

atmos_model.F90
rmontuoro: This block is not used when coupling with GOCART.

io/module_write_netcdf_parallel.F90
junwang-noaa: Do you know what is the issue?

atmos_model.F90
junwang-noaa: We may have issue for this with 32BIT=Y, but this can be fixed later.

io/module_write_netcdf_parallel.F90
rmontuoro: It was a divide by zero if `scale_fact = 0`

PR #318, opened 2021-05-25
Move IO related configuration from fv3_cap to write grid component

DusanJovic-NOAA:
## Description

Update fv3_cap's SetServices to use ONLY the semantic labels for each of the phases
Move IO related configuration from fv3_cap to write grid component
Code cleanup.

Is a change of answers expected from this PR?  No

## Testing

## Dependencies

https://github.com/ufs-community/ufs-weather-model/pull/600


junwang-noaa (approved):

rsdunlapiv (approved):

rsdunlapiv:
@DusanJovic-NOAA my recommendation would be to try adding just the INIT semantic labels and leave the RUN phases the same.  Basically, just work systematically adding a few labels at a time to see which one is causing the issue.  It would really be good to remove the IPDxPyy stuff as it is not intuitive - so its worth another pass if you have time.

DusanJovic-NOAA:
> @DusanJovic-NOAA my recommendation would be to try adding just the INIT semantic labels and leave the RUN phases the same. Basically, just work systematically adding a few labels at a time to see which one is causing the issue. It would really be good to remove the IPDxPyy stuff as it is not intuitive - so its worth another pass if you have time.

Yes. After this PR is merged, I'll go back and try to see where the issue is.

PR #321, opened 2021-05-28
inline post restart issue with gnu compiler

junwang-noaa:
## Description

This PR is to fix the issue with inline post in restart runs with gnu compiler. 
Is a change of answers expected from this PR?  The fix does not changes results. 

### Issue(s) addressed

- fixes noaa-emc/fv3atm/issues/<320>

## Testing

Run the control restart test with gnu compiler 

What compilers / HPCs was it tested with?  GNU
Are the changes covered by regression tests? 
NO, the current control restart does not have inline post. The global tests update (ufs-weather-model PR#561) has inline post turned on in the control and restart.
Tested on hera. Will test on other tier-1 platforms 
- Will the code updates change regression test baseline? 
This change does not change results, but the ufs-weather-model PR#561 does require new baseline.

## Dependencies

- [ ufs-weather-model/pull/<561>](https://github.com/ufs-community/ufs-weather-model/pull/561)
-  [noaa-emc/fv3atm/pull/<321>](https://github.com/NOAA-EMC/fv3atm/pull/321)

### DIFF
--- a/io/post_gfs.F90
+++ b/io/post_gfs.F90
@@ -158,9 +158,11 @@ module post_gfs
             filenameflat = 'postxconfig-NT.txt'
             if(size(paramset)>0) then
               do i=1,size(paramset)
-                if (size(paramset(i)%param)>0) then
-                  deallocate(paramset(i)%param)
-                  nullify(paramset(i)%param)
+                if (associated(paramset(i)%param)) then
+                  if (size(paramset(i)%param)>0) then
+                    deallocate(paramset(i)%param)
+                    nullify(paramset(i)%param)
+                  endif
                 endif
               enddo
               deallocate(paramset)
--- a/io/post_gfs.F90
+++ b/io/post_gfs.F90
@@ -156,7 +156,7 @@ module post_gfs
             if(mype==0) print *,'af read_xml at fh00,name=',trim(filenameflat)
           else if(ifhr > 0) then
             filenameflat = 'postxconfig-NT.txt'
-            if(size(paramset)>0) then
+            if(associated(paramset) .and. size(paramset)>0) then
               do i=1,size(paramset)
                 if (associated(paramset(i)%param)) then
                   if (size(paramset(i)%param)>0) then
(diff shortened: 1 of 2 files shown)


io/post_gfs.F90
```diff
@@ -156,11 +156,13 @@ subroutine post_run_gfs(wrt_int_state,mypei,mpicomp,lead_write,      &
             if(mype==0) print *,'af read_xml at fh00,name=',trim(filenameflat)
           else if(ifhr > 0) then
             filenameflat = 'postxconfig-NT.txt'
-            if(size(paramset)>0) then
+            if(associated(paramset) .and. size(paramset)>0) then
```
climbfuji: The Fortran compiler specifications do not state that the tests in an if clause will be executed ini the order written. You need to do the following to be fully compliant:
```
if (associated(paramset)) then
  if (size(paramset)>0) then
```

io/post_regional.F90
```diff
@@ -160,11 +160,13 @@ subroutine post_run_regional(wrt_int_state,mypei,mpicomp,lead_write,      &
             if(mype==0) print *,'af read_xml at fh00,name=',trim(filenameflat)
           else if(ifhr > 0) then
             filenameflat = 'postxconfig-NT.txt'
-            if(size(paramset)>0) then
+            if(associated(paramset) .and. size(paramset)>0) then
```
climbfuji: Same here

climbfuji (changes requested):

DusanJovic-NOAA (approved):

io/post_gfs.F90
junwang-noaa: Actually the size(paramset) is 1 but associated(paramset) is false with gnu compiler.

io/post_regional.F90
climbfuji: How? If `associated(paramset)` (means it's a pointer) is false, how can the size be 1?

junwang-noaa:
That is what I got when I tried with gnu compiler. The issue happened when
it tried to get the size of paramset(1)%param.

io/post_regional.F90
climbfuji: > That is what I got when I tried with gnu compiler. The issue happened when it tried to get the size of paramset(1)%param.
> […](#)

I think it would be best to make the same changes as in lines 161-162, 165-166.

climbfuji (approved):
Thanks!

PR #332, opened 2021-06-23
tracer tendency diagnostics and wrf-like per-timestep diagnostic

SamuelTrahanNOAA:
## Description

Adds tracer tendency diagnostics for all non-chemical tracers, replaces d*3dt variables with a more efficient storage format, and adds a WRF-like per-timestep surface pressure diagnostic.

### Issue(s) addressed

No issues were opened, as far as I know.

## Testing

> How were these changes tested?  

Eight CCPP suites were tested with all diagnostics enabled. This led to 30-90 variables, all of which were examined visually. A more quantitative comparison, comparing the accumulated change to the total change, proved to be difficult since 64-bit floating point is not sufficient for many processes. (That was confirmed experimentally.) However, these tests matched as much as could be expected, given the limitations of precision. Also, the full suite of regression tests were run to ensure non-diagnostic fields won't change.

> What compilers / HPCs was it tested with?

Hera with Intel and GNU. Jet with Intel until FV3 stopped building on Jet (again).
  
> Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)  

Yes. There are two new regression tests: control_diag_debug and fv3_gsd_diag_debug

> Have the ufs-weather-model regression test been run? On what platform?  

Yes. Hera Intel and Hera GNU.

> Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.

These code changes will not change the regression test baselines.

> Please commit the regression test log files in your ufs-weather-model branch

https://github.com/ufs-community/ufs-weather-model/pull/660

## Dependencies

> If testing this branch requires non-default branches in other repositories, list them.
> Those branches should have matching names (ideally)

ccpp-physics PR https://github.com/NCAR/ccpp-physics/pull/681
ufs-weather-model PR https://github.com/ufs-community/ufs-weather-model/pull/660

### DIFF
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -382,7 +382,9 @@ subroutine update_atmos_radiation_physics (Atmos)
 
     ! Per-timestep diagnostics must be after physics but before
     ! flagging the first timestep.
-    call atmos_timestep_diagnostics(Atmos)
+    if(GFS_control%print_diff_pgr) then
+      call atmos_timestep_diagnostics(Atmos)
+    endif
     
     ! Update flag for first time step of time integration
     GFS_control%first_time_step = .false.
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -227,8 +227,10 @@ contains
 
 subroutine update_atmos_radiation_physics (Atmos)
 !-----------------------------------------------------------------------
+  implicit none
   type (atmos_data_type), intent(in) :: Atmos
 !--- local variables---
+    integer :: idtend, itrac
     integer :: nb, jdat(8), rc, ierr
 
     if (mpp_pe() == mpp_root_pe() .and. debug) write(6,*) "statein driver"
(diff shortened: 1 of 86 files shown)


atmos_model.F90
```diff
@@ -359,6 +380,10 @@ subroutine update_atmos_radiation_physics (Atmos)
 
     endif
 
+    ! Per-timestep diagnostics must be after physics but before
+    ! flagging the first timestep.
+    call atmos_timestep_diagnostics(Atmos)
```
junwang-noaa: could this be added as an option?

climbfuji (approved):
Looks good to me. Just to confirm, did you add scheme `phys_tend` to the new suites that were committed to the repository in the last few PRs?

SamuelTrahanNOAA:
> Just to confirm, did you add scheme phys_tend to the new suites that were committed to the repository in the last few PRs

Yes, but it does nothing if ldiag3d=.false.

junwang-noaa (approved):
We need to point to ccpp physics main develop branch after ccpp PR is committed.

SamuelTrahanNOAA:
The ccpp/physics submodule now points to NCAR/main.

PR #337, opened 2021-07-01
update CCPP standard names for consistency

grantfirl:
## Description

This pull request only changes CCPP standard names in metadata files for those FV3 variables that are needed by all CCPP physics schemes in the current ccpp-physics repository. There are NO functional code changes and therefore no change in expected answers as a result of this pull request.

The standard names should better correspond to CF conventions (https://cfconventions.org/Data/cf-standard-names/77/build/cf-standard-name-table.html) and the superset of names represented in the CCPPStandardNames repository (https://github.com/ESCOMP/CCPPStandardNames/pull/13). A set of rules (which again relies heavily on CF naming conventions) is added to that same repository by https://github.com/ESCOMP/CCPPStandardNames/pull/14 and is effectively an addendum to the CF conventions (http://cfconventions.org/Data/cf-standard-names/docs/guidelines.html).

Note that this is only a partial update. Standard names for variables in the Diag and Interstitial derived data types as well as physical constants and other miscellaneous variables with metadata will be addressed in a future pull request.

### Issue(s) addressed

- partially fixes https://github.com/NCAR/ccpp-physics/issues/638

## Testing

These changes were tested as part of the existing set of UFS regression tests on Hera and CCPP-SCM regression tests.

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/687
- associated with https://github.com/ESCOMP/CCPPStandardNames/pull/13
- associated with https://github.com/NCAR/ccpp-scm/pull/269

### DIFF
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -10,14 +10,14 @@
   standard_name = geopotential_at_interface
   long_name = geopotential at model layer interfaces
   units = m2 s-2
-  dimensions = (horizontal_loop_extent,vertical_dimension_plus_one)
+  dimensions = (horizontal_loop_extent,vertical_interface_dimension)
   type = real
   kind = kind_phys
 [prsi]
   standard_name = air_pressure_at_interface
   long_name = air pressure at model layer interfaces
   units = Pa
-  dimensions = (horizontal_loop_extent,vertical_dimension_plus_one)
+  dimensions = (horizontal_loop_extent,vertical_interface_dimension)
   type = real
   kind = kind_phys
 [prsi(:,1)]
@@ -28,14 +28,14 @@
   type = real
   kind = kind_phys
 [prsik]
-  standard_name = dimensionless_exner_function_at_model_interfaces
+  standard_name = dimensionless_exner_function_at_interface
   long_name = dimensionless Exner function at model layer interfaces
   units = none
-  dimensions = (horizontal_loop_extent,vertical_dimension_plus_one)
+  dimensions = (horizontal_loop_extent,vertical_interface_dimension)
   type = real
   kind = kind_phys
 [prsik(:,1)]
-  standard_name = dimensionless_exner_function_at_lowest_model_interface
+  standard_name = surface_dimensionless_exner_function
   long_name = dimensionless Exner function at lowest model interface
   units = none
   dimensions = (horizontal_loop_extent)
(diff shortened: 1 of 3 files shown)


ccpp/data/GFS_typedefs.meta
```diff
@@ -28,14 +28,14 @@
...
   long_name = dimensionless Exner function at model layer interfaces
   units = none
-  dimensions = (horizontal_loop_extent,vertical_dimension_plus_one)
+  dimensions = (horizontal_loop_extent,vertical_interface_dimension)
   type = real
   kind = kind_phys
 [prsik(:,1)]
-  standard_name = dimensionless_exner_function_at_lowest_model_interface
+  standard_name = surface_dimensionless_exner_function
   long_name = dimensionless Exner function at lowest model interface
```
SMoorthi-emc, line 39: Exner function by definition is dimensionless.  So why add redundancy in the name?

ccpp/data/GFS_typedefs.meta
```diff
@@ -87,11 +87,11 @@
   standard_name = x_wind
   long_name = zonal wind
   units = m s-1
-  dimensions = (horizontal_loop_extent,vertical_dimension)
+  dimensions = (horizontal_loop_extent,vertical_layer_dimension)
   type = real
   kind = kind_phys
 [ugrs(:,1)]
-  standard_name = x_wind_at_lowest_model_layer
+  standard_name = x_wind_at_surface_adjacent_layer
```
SMoorthi-emc, line 94: Why not "zonal" or "eastward"?

ccpp/data/GFS_typedefs.meta
grantfirl: CF has "dimensionless_exner_function" so we are more-or-less forced to adopt this if our standard names are to be a superset of theirs. This is a good question that we plan to ask the CF standards committee, however, and we had the same question. In fact, we've collected several questions for them regarding their names, but have not tried to get anything changed yet.

ccpp/data/GFS_typedefs.meta
grantfirl: Another good question. I think that the "x_wind*" names are a nod to different grids, and [x,y]_wind were deemed to be more general (work for a greater number of physics and hosts) than [eastward,northward]_wind. But, [x,y]_wind would also be deficient for, say, irregular or hexagonal grids too, so I'm not sure how valuable sticking with [x,y] really is.

CF standards do have [eastward,northward]_wind that we can use, though. If I remember correctly, FV3 converts winds on their C/D grid to true zonal/meridional winds on the A grid for the physics to use, correct? So, changing to [eastward,northward] should work in that case.

grantfirl: See https://github.com/NCAR/ccpp-physics/discussions/692 for more discussion.

junwang-noaa:
This PR is committed in PR#356, close PR.

Issue #253, opened 2021-03-04
Need a better way to pass global communicator (w/o I/O tasks) to physics


climbfuji:
## Description
When the CCPP physics were introduced, it was required to pass the global communicator that includes all compute tasks, but not the I/O tasks, to the physics for the purpose of broadcasting lookup tables at physics init time etc.

This was implemented by @climbfuji by exporting `commglobal` from the dycore to populate a member of the GFS control DDT. As @bensonr pointed out, we should be able to get this communicator directly from ESMF (which passes it to the dycore in the first place).

## Solution

Pass the global communicator that includes all compute tasks, but not the I/O tasks, from ESMF directly to `atmos_model.F90` and stop exporting `commglobal` from the dycore.

## Alternatives

Keep the current, inelegant solution.

## Testing:

n/a

## Dependent PRs:

n/a

junwang-noaa:
@bensonr the forecast tasks communicator is passed to fms_init in the code [here](https://github.com/NOAA-EMC/fv3atm/blob/develop/module_fcst_grid_comp.F90#L246), can we get the communicator out from mpp or fms functions?

bensonr:
@junwang-noaa - we could, but I think the communicator should be pushed
down to atmos_model_init as an argument.  It should then be included as an
element of the GFS_init_type and sent down to the physics via the Init_parm
ddt - and not as yet another argument to GFS_Initialize (commglobal)

At the same time, the argument to GFS_Initialize being satisfied via
mpp_npes() should also become an element in GFS_init_type.  The whole
reason for this particular ddt is to keep from having to make interface
changes as you find another variable you'd like to pass down into another
component or sub-component, in this case the physics initialization.  If
this is allowed to continue, you could quickly find yourself back to the
olden days of 100+ arguments going between components and sub-components.

junwang-noaa:
Code changes are committed in PR#345, close issue.

Issue #136, opened 2020-07-01
check fdiag consistency with fhmax/FHOUT/FHOUT_HF


junwang-noaa:
In current fv3, the fdiag is defined in namelist while FHMAX, FHOUT/FHOUT_HF are defined in the model configuration. there could be difference in fdiag and fhout/fhout_hf which can problems such as model output does not change. This ticket is to fix the issue to make sure the fdiag and fhout/fhou_hf to be consistent.

junwang-noaa:
Code changes are committed in PR#345, close issue.

PR #345, opened 2021-07-15
Flexible output hour for history files

junwang-noaa:
## Description

In this PR, model is allowed to output history files at flexible (random) forecast hours. 
-   fdiag, fhout,fhmax, fhouthf, and fhmaxhf are removed from fv3 namelist atmos_model_nml. With the code updates, those variables should be deleted from namelist.
-  in model_configure, there are two ways to control output forecast time
  i.  Use nfhout, nfhmax, nfhout_hf, nfhmax_hf, nsout, the meanings of those variables do not change.
 ii.  Use output_fh to control output time:
    a) An array of output_fh means to output history files at those forecast time, e.g.:
    "output_fh: 0 1 2 3 12 24 48"
    If the first element is zero, it means the first time step output.
    b) If there are two elements in output_fh and the second one has value -1, it means the first element is output frequency. e.g.
    " output_fh: 6 -1"
    fv3atm will output the history file every 6 hours.
- A related change for  restart_interval that controls when to write out restart files
 a) An array of restart_interval means to output restart files at those forecast time. If there is one element, it means to write to at that single forecast time (forecast time is from the initial start time, not restart time). 
 b) to use restart frequency, two elements need to be specified, and the second element is "-1"
- Pass forecast mpi communicator to Init_parm in atmos_model_init

### Issue(s) addressed

- fixes fv3atm issue [#136](https://github.com/NOAA-EMC/fv3atm/issues/136)
- fixes fv3atm issue [#253](https://github.com/NOAA-EMC/fv3atm/issues/253)
- ufs-weather-model issue [#339](https://github.com/ufs-community/ufs-weather-model/issues/339)

## Testing

- The code has been tested on hera. Will merge when the develop branch is updated. 
- No baseline results change

## Dependencies

-fv3atm PR [#345](https://github.com/NOAA-EMC/fv3atm/pull/345)
-ufs-weather-model PR [#691](https://github.com/ufs-community/ufs-weather-model/pull/691)


junwang-noaa:
@climbfuji The forecast grid comp communicator fcst_mpi_comm is passed to GFS_init_type Init_parm in atmos_model_init. I do not use global comm as global may have different meaning in different context, please let me know if you want to change it. Let me know if this resolves the issue [#253](https://github.com/NOAA-EMC/fv3atm/issues/253)

climbfuji:
> @climbfuji The forecast grid comp communicator fcst_mpi_comm is passed to GFS_init_type Init_parm in atmos_model_init. I do not use global comm as global may have different meaning in different context, please let me know if you want to change it. Let me know if this resolves the issue [#253](https://github.com/NOAA-EMC/fv3atm/issues/253)

I think this should work, yes. We should then remove `use fv_mp_mod, only: commglobal` and replace all occurrences of `commglobal` with `fcst_mpi_comm` in `atmos_model.F90`. Thanks for taking care of this!

DusanJovic-NOAA:
If forecast grid comp communicator is passed down to ccpp via `Init_parm` derived type, then `commglobal` argument in `GFS_initialize` is redundant, and can be confusing and potentially inconsistent. Should be removed.

junwang-noaa:
> 
> 
> If forecast grid comp communicator is passed down to ccpp via `Init_parm` derived type, then `commglobal` argument in `GFS_initialize` is redundant, and can be confusing and potentially inconsistent. Should be removed.

Done

module_fv3_config.F90
```diff
@@ -25,6 +26,7 @@ module module_fv3_config
   logical                  :: force_date_from_configure
   logical                  :: restart_endfcst
 !
+  real,dimension(:),allocatable                   :: output_fh
```
DusanJovic-NOAA, line 29: I think this array (`output_fh`) should be an integer array storing seconds, not real array storing hours. The elements of this array are compared for equality to `nfseconds` or `seconds`, and both of them are integers. Then there will be no need to have multiplication by 3600 or division by 3600 at so many places. As soon as configuration parameters `nfhout`, `nfhmax_hf` and  `nfhout_hf` are read in they can be converted to seconds and all arithmetic should be done with integers (seconds) only. Which will simplify the code.

fv3_cap.F90
```diff
@@ -1041,47 +1095,19 @@ subroutine ModelAdvance_phase2(gcomp, rc)
 
     time_elapsed  = currtime - starttime
     na = nint(time_elapsed/timeStep)
-
-    ! if(mype==0) print *,'in fv3_cap,in model run, advance2,na=',na
+    call ESMF_TimeIntervalGet(time_elapsed, h=nfh, m=nfm, s=nfs, rc=rc)
+    nfseconds = nfh*3600.+ nfm*60. + nfs
```
DusanJovic-NOAA: `3600.` should be just `3600`, `60.` should be `60`, since `nfseconds`, `nfh`, `nfm` and `nfs` are all integers. No need to have the RHS expression implicitly converted to real, and then converted back to integer.

DusanJovic-NOAA: I think ESMF_TimeIntervalGet can be called like:

`    call ESMF_TimeIntervalGet(time_elapsed, s=nfseconds, rc=rc)
`

to return the seconds directly.

atmos_model.F90
```diff
@@ -157,9 +160,8 @@ module atmos_model_mod
 !logical :: debug        = .true.
 logical :: sync         = .false.
 integer, parameter     :: maxhr = 4096
```
DusanJovic-NOAA: `maxhr` is not used anymore. Should be deleted.

junwang-noaa: done

fv3_cap.F90
junwang-noaa: Yes. it's good to use s=nfseconds, Done.

module_fv3_config.F90
junwang-noaa: For history reasons, this array is storing forecast hours (especially for climate runs) to maintain readability. In the code, so far we are comparing forecast time in seconds. But In the future, if finer time available, this can be extended to compare time in real seconds. So I think it's good to keep output_fh for forecast hour as real number, not integer seconds

DusanJovic-NOAA (approved):

climbfuji (approved):
Thanks for fixing the MPI communicator issue!

PR #349, opened 2021-07-16
Fix time dimension in restart files

### LINKED ISSUE #344: dimension change in sfc restart files
## Description
This is reported by Bin Liu.

After syncing with the ufs-weather-model develop as of 07/12/2012. I notice there is a change in the model output sfc_data restart files.

It looks to me there was a change of the "Time" dimension size from 1 into 2, and also only the zorl variable has the Time dimension now. 
      double tg3 ( yaxis_1, xaxis_1 )
         long_name :    tg3
         units :        none
         checksum :     90839D8A4C3CC0C1

      double zorl ( Time, yaxis_1, xaxis_1 )
         long_name :    zorl
         units :        none
         checksum :     471AD51A1F40C999,EF5F56B3B8D8CB1A

In contrast, previously, all the surface variables have the Time dimension and the Time dimension size is 1.
      double tg3 ( Time, yaxis_1, xaxis_1 )
         long_name :    tg3
         units :        none
         checksum :     F9DF63348B3CECF2

      double zorl ( Time, yaxis_1, xaxis_1 )
         long_name :    zorl
         units :        none
         checksum :     7B1F22BE4A631212

Not sure if this is an expected change. If so, could you point us to the related background/context for this change? 

Also, I checked that in the new sfc_data.nc restart file, for the zorl variable, the first dimension is actually not really Time. zorl(1,:,:) is actually zorlo (which has missing value on land), and zorl(2,:,:) has actual zorl value for both land and water points.

      double zorl ( Time, yaxis_1, xaxis_1 )
         long_name :    zorl
         units :        none
         checksum :     471AD51A1F40C999,EF5F56B3B8D8CB1A

So, this might be introduced improperly. A better solution would be adding the zorlo variable in the sfc_data.nc (since there are zorli, zorll variables already). Also, Time dimension should be restored properly as in other fv_core and phy_data files.

### To Reproduce:
What compilers/machines are you seeing this with? on all the platforms
1. run RT test and check the RESTART/sfc_data.tile*.nc file

### LINKED ISSUE #348: Wrong name of variable in FV3GFS_io.F90: 'weasd' should be 'weasdl'
Line 1568 in `FV3GFS_io.F90` contains an error:

https://github.com/NOAA-EMC/fv3atm/blob/33c1a9c172f96f3117257e5d416529671fc38c61/io/FV3GFS_io.F90#L1568

This line should be 'weasdl' and not 'weasd'.

This bug will be corrected at the same time as https://github.com/NOAA-EMC/fv3atm/issues/344.

climbfuji:
## Description

- Fix bug in `io/FV3GFS_io.F90` to prevent the time dimension having length 2: read/write `zorl` into/from `Sfcprop%zorl` instead of `Sfcprop%zorlw`. Adjust logic to compute optiional `zorl*` variables.
- Fix bug in `io/FV3GFS_io.F90`: replace `weasd` with `weasdl` in line 1568.
- Rearrange optional fields (indices 32 to 39) in `io/FV3GFS_io.F90`

### Issue(s) addressed

Fixes https://github.com/NOAA-EMC/fv3atm/issues/344
Fixes https://github.com/NOAA-EMC/fv3atm/issues/348

## Testing

See MISSING

## Dependencies

MISSING

### DIFF
--- a/io/FV3GFS_io.F90
+++ b/io/FV3GFS_io.F90
@@ -784,13 +784,13 @@ module FV3GFS_io_mod
       sfc_name2(31) = 'snoalb'
       !--- variables below here are optional
       sfc_name2(32) = 'sncovr'
-      sfc_name2(33) = 'tsfcl' !temp on land portion of a cell
-      sfc_name2(34) = 'zorll' !zorl on land portion of a cell
-      sfc_name2(35) = 'zorli' !zorl on ice portion of a cell
-      sfc_name2(36) = 'snodl' !snowd on land portion of a cell
-      sfc_name2(37) = 'tsfc'  !tsfc composite
-      sfc_name2(38) = 'zorlw' !zorl on water portion of a cell
-      sfc_name2(39) = 'weasdl'!weasd on land portion of a cell
+      sfc_name2(33) = 'snodl' !snowd on land portion of a cell
+      sfc_name2(34) = 'weasdl'!weasd on land portion of a cell
+      sfc_name2(35) = 'tsfc'  !tsfc composite
+      sfc_name2(36) = 'tsfcl' !temp on land portion of a cell
+      sfc_name2(37) = 'zorlw' !zorl on water portion of a cell
+      sfc_name2(38) = 'zorll' !zorl on land portion of a cell
+      sfc_name2(39) = 'zorli' !zorl on ice portion of a cell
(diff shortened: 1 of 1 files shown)


climbfuji:
Merged as part of https://github.com/NOAA-EMC/fv3atm/pull/350.

PR #388, opened 2021-09-15
Convert real(kind_phys) vegetation, slope and soil type arrays into integer arrays without affecting input/output files

climbfuji:
## Description

This PR converts the real vegetation, slope and soil type arrays and replaces them with the existing (interstitial) "integer" arrays (these become persistent). Additional `_save` variables are added as required for obtaining b4b identical output (see https://github.com/NCAR/ccpp-physics/pull/730).

The approach taken here was to take the conversion from integer to real for writing the data to disk to the latest point possible, and to convert back from real to integer when reading from disk. This ensures that there are zero changes to upstream and downstream processes. It should also be noted that the I/O layer currently does not support writing/reading integer data and that, if changes to the input/output files are preferred in a follow-up PR, additional development will be necessary.

### Issue(s) addressed

- fixes https://github.com/NCAR/ccpp-physics/issues/704

## Testing

For regression testing, see https://github.com/ufs-community/ufs-weather-model/pull/804

## Dependencies

https://github.com/NCAR/ccpp-physics/pull/730
https://github.com/NOAA-EMC/fv3atm/pull/388
https://github.com/ufs-community/ufs-weather-model/pull/804


io/FV3GFS_io.F90
```diff
@@ -2638,101 +2638,122 @@ subroutine fv3gfs_diag_output(time, diag, atm_block, nx, ny, levs, ntcw, ntoz, &
...
-             !--- need to "mask" gflux to output valid data over land/ice only
-             var2(1:nx,1:ny) = missing_value
-             do j = 1, ny
-               jj = j + jsc -1
-               do i = 1, nx
-                 ii = i + isc -1
-                 nb = Atm_block%blkno(ii,jj)
-                 ix = Atm_block%ixp(ii,jj)
-                  if (Diag(idx)%data(nb)%var21(ix) /= 0) var2(i,j) = Diag(idx)%data(nb)%var2(ix)*lcnvfac
+               call store_data(Diag(idx)%id, var2, Time, idx, Diag(idx)%intpl_method, Diag(idx)%name)
```
climbfuji, line 2654: This GitHub diff view is confusing. The only thing that changed was adding the `int_or_real: if` case and move the existing code into the `else` block.

junwang-noaa: Maybe in the future we need to consider outputting those fields as integer too.

climbfuji: Yes, that would be nice, but it will require changes to the pre- and post-processing chain.

junwang-noaa (approved):

DusanJovic-NOAA (approved):

climbfuji:
This PR is ready to merge, ccpp-physics hash is up to date.

PR #396, opened 2021-09-23
updates associated with CA restart and PE decomposition bugs

pjpegion:
## Description

Provide a detailed description of what this PR does.  
What bug does it fix, or what feature does it add?  
Is a change of answers expected from this PR?  

### Issue(s) addressed

- fixes #<issue_number>
- fixes noaa-emc/fv3atm/issues/<issue_number>
## Testing

  Restart and decompostion reproducbility were tested using the regression test setup.
  gnu and intel, mostly on hera.
  A new regression test (control_ca_restart) will be added.
 A limited set of RTs were run on hera
-   Yes a new baseline will need to be generated

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/737
 associated PR 
- https://github.com/noaa-psd/stochastic_physics/pull/47

### DIFF
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -699,8 +699,8 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
                               GFS_data%Coupling, GFS_data%Grid, GFS_data%Tbd, GFS_data%Cldprop,  GFS_data%Radtend, &
                               GFS_data%IntDiag, Init_parm, GFS_Diag)
    call FV3GFS_restart_read (GFS_data, GFS_restart_var, Atm_block, GFS_control, Atmos%domain, Atm(mygrid)%flagstruct%warm_start)
-   if(GFS_control%ca_sgs)then
-      call read_ca_restart (Atmos%domain,GFS_control%scells)
+   if(GFS_control%do_ca)then
+      call read_ca_restart (Atmos%domain,GFS_control%scells,GFS_control%nca,GFS_control%nca_g)
    endif
    ! Populate the GFS_data%Statein container with the prognostic state
    ! in Atm_block, which contains the initial conditions/restart data.
@@ -993,8 +993,8 @@ subroutine atmos_model_end (Atmos)
         GFS_Control%lndp_type > 0  .or. GFS_Control%do_ca ) then
       if(restart_endfcst) then
         call write_stoch_restart_atm('RESTART/atm_stoch.res.nc')
-        if (GFS_control%ca_sgs)then
-          call write_ca_restart(Atmos%domain,GFS_control%scells)
+        if (GFS_control%do_ca)then
+          call write_ca_restart()
         endif
       endif
       call stochastic_physics_wrapper_end(GFS_control)
@@ -1022,8 +1022,8 @@ subroutine atmos_model_restart(Atmos, timestamp)
     call atmosphere_restart(timestamp)
     call FV3GFS_restart_write (GFS_data, GFS_restart_var, Atm_block, &
                                GFS_control, Atmos%domain, timestamp)
-    if(GFS_control%ca_sgs)then
-       call write_ca_restart(Atmos%domain,GFS_control%scells,timestamp)
+    if(GFS_control%do_ca)then
+       call write_ca_restart(timestamp)
     endif
 end subroutine atmos_model_restart
 ! </SUBROUTINE>
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -699,8 +699,8 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
                               GFS_data%Coupling, GFS_data%Grid, GFS_data%Tbd, GFS_data%Cldprop,  GFS_data%Radtend, &
                               GFS_data%IntDiag, Init_parm, GFS_Diag)
    call FV3GFS_restart_read (GFS_data, GFS_restart_var, Atm_block, GFS_control, Atmos%domain, Atm(mygrid)%flagstruct%warm_start)
-   if(GFS_control%do_ca)then
-      call read_ca_restart (Atmos%domain,GFS_control%scells,GFS_control%nca,GFS_control%nca_g)
+   if(GFS_control%do_ca .and. Atm(mygrid)%flagstruct%warm_start)then
+      call read_ca_restart (Atmos%domain,GFS_control%scells,GFS_control%nca,GFS_control%ncells_g,GFS_control%nca_g)
    endif
    ! Populate the GFS_data%Statein container with the prognostic state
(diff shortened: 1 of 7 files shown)


ccpp/data/GFS_typedefs.F90
```diff
@@ -1122,7 +1121,7 @@ module GFS_typedefs
     logical              :: ca_sgs          !< switch for sgs ca
     logical              :: ca_global       !< switch for global ca
     logical              :: ca_smooth       !< switch for gaussian spatial filter
-    integer              :: iseed_ca        !< seed for random number generation in ca scheme
+    integer*8            :: iseed_ca        !< seed for random number generation in ca scheme
```
climbfuji: I think we need to be careful here. What is the type / kind of this variable in `GFS_typedefs.meta`? I think we should use a proper kind variable and, if this variable ever gets used in CCPP, we should set type and kind accordingly in the metadata, too. Maybe we should do this right away to not forget about it? It doesn't cause a problem right now the way it is written. What do you think?

ccpp/data/GFS_typedefs.F90
```diff
@@ -3053,7 +3051,7 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
                                  nwat, tracer_names, tracer_types,  &
                                  input_nml_file, tile_num, blksz,   &
                                  ak, bk, restart, hydrostatic,      &
-                                 communicator, ntasks, nthreads)
+                                 communicator, ntasks, nthreads )
```
climbfuji: Don't think this change is necessary?

pjpegion: Hmm, I reverted that change.  Guess I need to do it again.

ccpp/data/GFS_typedefs.F90
pjpegion: I switched this to kind_dbl_prec in GFS_typdefs, and in stochastic_physisics

climbfuji (approved):

atmos_model.F90
```diff
@@ -685,8 +685,8 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
                               GFS_data%Coupling, GFS_data%Grid, GFS_data%Tbd, GFS_data%Cldprop,  GFS_data%Radtend, &
                               GFS_data%IntDiag, Init_parm, GFS_Diag)
    call FV3GFS_restart_read (GFS_data, GFS_restart_var, Atm_block, GFS_control, Atmos%domain, Atm(mygrid)%flagstruct%warm_start)
-   if(GFS_control%ca_sgs)then
-      call read_ca_restart (Atmos%domain,GFS_control%scells)
+   if(GFS_control%do_ca .and. Atm(mygrid)%flagstruct%warm_start)then
+      call read_ca_restart (Atmos%domain,GFS_control%scells,GFS_control%nca,GFS_control%ncells_g,GFS_control%nca_g)
    endif
```
lisa-bengtsson, line 690: @pjpegion per our previous conversation, what happens if you warm start the run and the CA files do not exist? Do you cold start the CA or abort? Should the flag instead read "restart" instead of "warm_start"?

stochastic_physics/stochastic_physics_wrapper.F90
```diff
@@ -317,61 +304,37 @@ subroutine stochastic_physics_wrapper (GFS_Control, GFS_Data, Atm_block, ierr)
...
-            GFS_Control%ca_smooth,GFS_Control%nspinup,GFS_Control%ca_trigger,Atm_block%blksz(1),GFS_Control%master,GFS_Control%communicator)
+            GFS_Control%nthresh,GFS_Control%rcell,GFS_Control%tile_num,GFS_Control%nca,GFS_Control%scells,GFS_Control%tlives,     &
+            GFS_Control%nfracseed, GFS_Control%nseed,GFS_Control%iseed_ca,                                                        &
+            GFS_Control%nspinup,GFS_Control%ca_trigger,Atm_block%blksz(1),GFS_Control%master,GFS_Control%communicator)
          ! Copy contiguous data back as needed
          do nb=1,nblks
-             GFS_Data(nb)%Intdiag%ca_deep(:)  = ca_deep_diag(nb,1:GFS_Control%blksz(nb))
-             GFS_Data(nb)%Intdiag%ca_turb(:)  = ca_turb_diag(nb,1:GFS_Control%blksz(nb))
-             GFS_Data(nb)%Intdiag%ca_shal(:)  = ca_shal_diag(nb,1:GFS_Control%blksz(nb))
+             GFS_Data(nb)%Intdiag%ca_deep(:)  = ca_deep_cpl(nb,1:GFS_Control%blksz(nb))
```
climbfuji: Wondering, why do we need to copy `ca_deep_cpl` into both `GFS_Data(nb)%Intdiag%ca_deep` and `GFS_Data(nb)%Coupling%ca_deep`? Do we need to have two copies? Same for `ca_shal` and `ca_turb`.

atmos_model.F90
pjpegion: in the case you describe, it will cold start the CA since update_ca.F90 again checks for the existence of the restart files, and warm_start is the correct flag.

stochastic_physics/stochastic_physics_wrapper.F90
pjpegion: We have that so we can output the ca fields using the diag_table.  Is not no longer necessary?

lisa-bengtsson: It's based on rather old code, from 2017, at the time the fields in GFS_diagnostics.F90 needed to be allocated in the "Diag" container, whereas Coupling fields were in the "Coupling" container defined in GFS_typedefs.F90. They are the same fields, just that one is coupled to the physics and one is used for diagnostics output as defined by diag_table. Perhaps there is a better way of doing this now?

climbfuji: I think those coupling fields can be sent directly to the diagnostic output.

pjpegion: I just realized that I didn't address this code change.  Just so you know, this is not an increase in the memory footprint of the code compared to what is in develop. I just moved to copies out of the CA code.

climbfuji: That's ok, I think there are a few more of those duplicate arrays that we should clean up in the near future.

pjpegion: I just did a clean up of the ca and sppt,shum,skeb etc. duplicates.  If it works, I will commit it.

junwang-noaa (approved):

climbfuji:
Submodule pointer for ccpp-physics is correct.

Issue #403, opened 2021-09-29
fix the output file name with non-integer output time and the diag_time for random output time


junwang-noaa:
## Description
The output file name does not have correct time stamp when the output time is not a integer. E.g. output_fh=0.25, the file name is currently atmf00.nc and sfcf00.nc. Also for random output time, the diag_time may not be updated when the output time is not multiple of fhzero. Fixed are provided to fix the two issues.

### To Reproduce:
1. check out the develop branch and run control test
2. change the output_fh in model_configure to 0.25, then run the test again. The file name does not contain the minutes, so they do not reflect correct output time.
3. after 1, change output_fh to 21, the accumulated fields in the model history files are different.

## Testing:

1. Have you tested the code changes? On what platforms?
Yes, on hera. 
2. Have you run regression test in ufs-weather-model or ufs-s2s-model with code changes?
- Will the baseline results change? No

## Dependent PRs:

- ufs-weather-model PR 
- fv3atm PR

junwang-noaa:
Code is merged, close issue.

PR #404, opened 2021-09-29
fix the output file name and diag time

junwang-noaa:
## Description

The output file name does not have correct time stamp when the output time is not a integer. E.g. output_fh=0.25, the file name is currently atmf00.nc and sfcf00.nc. Also for random output time, the diag_time may not be updated when the output time is not multiple of fhzero. This PR contains fixes for the two issues.

### Issue(s) addressed

- fv3atm issue [#403](https://github.com/NOAA-EMC/fv3atm/issues/403)

## Testing

How were these changes tested?  Tested on hera. 
Yes on hera.
- No

## Dependencies

ufs-weather-model PR
fv3atm PR


climbfuji (commented):
I don't understand the logic, especially in lines 763 to 772. Is it possible to document that code? Just a brief one liner what each of those blocks does would be great.

DusanJovic-NOAA:
Why do need all this logic with `lflname_fulltime`, why can we just switch to using the filename format that always include HH-MM-SS, which will simplify the code substantially. Scripting will be also simplified by always using the same file naming scheme, that does not change based on the value `output_fh`.

junwang-noaa:
> 
> 
> Why do need all this logic with `lflname_fulltime`, why can we just switch to using the filename format that always include HH-MM-SS, which will simplify the code substantially. Scripting will be also simplified by always using the same file naming scheme, that does not change based on the value `output_fh`.

This is for backward compatible. Otherwise all the downstream jobs will be broken.

DusanJovic-NOAA:
> > Why do need all this logic with `lflname_fulltime`, why can we just switch to using the filename format that always include HH-MM-SS, which will simplify the code substantially. Scripting will be also simplified by always using the same file naming scheme, that does not change based on the value `output_fh`.
> 
> This is for backward compatible. Otherwise all the downstream jobs will be broken.

Can't we announce the change and downstream jobs can update scripts. This is after all develop branch, we should be able to make breaking changes. Are we going to keep backward compatibility forever. Every change we make in any of the config file structure or output is potential breaking change for somebody.

fv3_cap.F90
```diff
@@ -746,6 +746,9 @@ subroutine InitializeAdvertise(gcomp, rc)
               endif
               do i=2,nfh
                 output_fh(i) = (i-1)*outputfh2(1) + output_startfh
+                ! Except fh000, which is the first time output, if any other of the 
+                ! output time is not integer hour, set lflname_fulltime to be ture, so the
```
climbfuji:
```suggestion
                ! output time is not integer hour, set lflname_fulltime to be true, so the 
```

fv3_cap.F90
```diff
@@ -760,6 +763,10 @@ subroutine InitializeAdvertise(gcomp, rc)
              count=noutput_fh, rc=rc)
           if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
           if( output_startfh == 0) then
+            ! if the output time in output_fh array contains first time stamp output,
```
climbfuji:
```suggestion
            ! If the output time in output_fh array contains first time stamp output,
```

fv3_cap.F90
```diff
@@ -760,6 +763,10 @@ subroutine InitializeAdvertise(gcomp, rc)
...
+            ! check the rest of output time, otherwise, check all the output time. 
+            ! if any of them is not integer hour, the history file names will
```
climbfuji:
```suggestion
            ! If any of them is not integer hour, the history file names will
```

fv3_cap.F90
```diff
@@ -773,6 +780,9 @@ subroutine InitializeAdvertise(gcomp, rc)
           else
             do i=1,noutput_fh
               output_fh(i) = output_startfh + output_fh(i)
+              ! when output_startfh >0, check all the output time, if any of
```
climbfuji:
```suggestion
              ! When output_startfh >0, check all the output time, if any of
```

fv3_cap.F90
```diff
@@ -773,6 +780,9 @@ subroutine InitializeAdvertise(gcomp, rc)
...
+              ! them is not integer hour, set lflname_fulltime to be ture. The
```
climbfuji:
```suggestion
              ! them is not integer hour, set lflname_fulltime to be true. The 
```

climbfuji (changes requested):
Thanks for adding those comments. A few suggestions/corrections of typos, otherwise looks good.

junwang-noaa:
@climbfuji I added some comments to explain, please check. Basically for frequency output, we will don't count the  the first time step output as it is labeled as fh000.

climbfuji:
> @climbfuji I added some comments to explain, please check. Basically for frequency output, we will don't count the the first time step output as it is labeled as fh000.

I did, please look at/accept my suggestions.

junwang-noaa:
@DusanJovic-NOAA Well, this will impact future operational implementations. As we don't know if we will have higher frequency output, I think we may need to get the downstream involved before we make the hard code change. Right now if the output time is non-integer (needs min/sec info), the output file will have full time stamp, otherwise if it's all integer hours, file names will only contain hours.

DusanJovic-NOAA (approved):

climbfuji (approved):

DeniseWorthen (approved):

PR #406, opened 2021-10-04
Properly import coupling fields when running with separate run phases

### LINKED ISSUE #401: Replacing single NUOPC run phase with two sequential run phases leads to surface layer differences in P7 configurations
## Description
Large differences in surface layer quantities occur in fully coupled (P7) experiments when running the atmosphere component (ATM) using two sequential run phases (phase1, phase2) instead of a single one. Such differences appear after the first coupling time step, and are most evident for surface temperature, 2m temperature, and surface heat fluxes.

### To Reproduce:
This issue was observed on Hera using Intel compilers.

To reproduce, please edit the UFS weather model's run sequence in the `nems.configure` file and replace `ATM` as follows:

```
[...]
ATM phase1
ATM phase2
[...]
```

rmontuoro:
## Description

This PR addresses a coupling issue described in #401. 

The `fv3_checkimport` method is now called also for run phase 1. The method has been refactored to properly check the timestamp of imported fields using updated ESMF API.

### Issue(s) addressed

- fixes #401 

## Testing

The changes were tested on the Hera/Intel platform.

## Dependencies

None


fv3_cap.F90
```diff
@@ -140,7 +140,7 @@ subroutine SetServices(gcomp, rc)
 
     ! specializations required to support 'inline' run sequences
     call NUOPC_CompSpecialize(gcomp, specLabel=label_CheckImport, &
-                              specPhaseLabel="phase1", specRoutine=NUOPC_NoOp, rc=rc)
+                              specPhaseLabel="phase1", specRoutine=fv3_checkimport, rc=rc)
```
junwang-noaa, line 143: So now the fv3_checkImport will be called once for one integration step with two phases, just as it is called once in "ModelAdvance". One question is: should we also call checkImport for phase2? or maybe it is no necessary as the time stamp does not change?

fv3_cap.F90
```diff
@@ -1250,25 +1250,32 @@ subroutine fv3_checkimport(gcomp, rc)
...
-            !TODO: introduce and use INCOMPATIBILITY return codes!!!!
+        if (isValid) then
+          ! check if field is current
+          isValid = NUOPC_IsAtTime(fieldList(n), currTime, rc=rc)
+          if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
+          if (.not.isValid) then
             call ESMF_LogSetError(ESMF_RC_ARG_BAD, &
-                                  msg="NUOPC INCOMPATIBILITY DETECTED: Import Field not at current time", &
+                                  msg="NUOPC INCOMPATIBILITY DETECTED: Import Field " &
+                                      // trim(fldname) // " not at current time", &
```
junwang-noaa, line 1274: Should we have: importFieldsValid(nf) = .false. here too? Or maybe we are now allowing import fields with different time stamp (not at currTime)?

fv3_cap.F90
rmontuoro: `fv3_checkimport` should be called at the beginning of the atmosphere's integration step for coupling with external components except GOCART. It is not called before run phase 2 since: (a) the timestamp of imported fields does not change, and (b) GOCART does not require it.

fv3_cap.F90
rmontuoro: We could set `importFieldsValid(nf) = .false.` here as well if we are not planning to abort when a time incompatibility is detected. Currently, we return an ESMF error code that will cause NUOPC to abort.

junwang-noaa (approved):

fv3_cap.F90
```diff
@@ -1212,7 +1212,7 @@ subroutine fv3_checkimport(gcomp, rc)
     type(ESMF_Clock)           :: clock
     type(ESMF_Time)            :: currTime, invalidTime
     type(ESMF_State)           :: importState
-    logical                    :: timeCheck1,timeCheck2
+    logical                    :: isValid
```
DusanJovic-NOAA, line 1215: This local variable `isValid` is used at three places and have three different meanings at all those places. I find it very confusing.
First it is used to indicate whether a field's timestamp is "a valid timestamp", from NUOPC's perspective, and this is the only place this variable should be used with this name. 
Second it is reused to check whether a field is not at "invalid time", where an invalid time is defined locally and set to an arbitrary/special value (year  99999999). Why is this necessary? How is this value (99999999) determined?
And finally `isValid` is reused for a third time to check whether a field is at current time. 
If all these checks are necessary, then three logical variables with different names should be used. For example  `isValid`, `isNotSpecialValue` and `isAtCurrentTime`, which will improve readability of the code at expense of adding two logical variables (8 bytes). 
I wonder if all these check can be combined in just one check that checks whether or not field's time is at current time?

rmontuoro: We could use multiple logical variables if needed for clarity. We could also write a single logical function combining all the tests for increased modularity. I don't believe such a function is required elsewhere in the code base, though, and it could lead to confusion in cases where a field timestamp has not been set at all.

DusanJovic-NOAA (approved):

PR #421, opened 2021-11-09
Use upp as a submodule

### LINKED ISSUE #905: Use UPP as a submodule
UPP library is updated very frequently, which requires making a UPP release, building it using hpc-stack on all supported platforms and then updating modules in ufs-weather-model. We can just add the upp as a submodule in fv3atm repo and update that submodule whenever we need to update inline post.

DusanJovic-NOAA:
## Description

This PR adds UPP submodule. It is used in inline post instead of an external upp library.
No change in baseline is expected.

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/905

## Testing

How were these changes tested?  Regression tested on Hera.
What compilers / HPCs was it tested with?  Intel.
Are the changes covered by regression tests? Yes. All tests that run inline-post.
Have the ufs-weather-model regression test been run? Yes. On what platform?  Hera.
- Will the code updates change regression test baseline? No.

## Dependencies

- waiting on https://github.com/NOAA-EMC/UPP/pull/412

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,6 @@
   path = ccpp/physics
   url = https://github.com/NCAR/ccpp-physics
   branch = main
+[submodule "upp"]
+  path = upp
+  url = https://github.com/DusanJovic-NOAA/UPP
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,4 +12,5 @@
   branch = main
 [submodule "upp"]
   path = upp
-  url = https://github.com/DusanJovic-NOAA/UPP
+  url = https://github.com/NOAA-EMC/UPP
+  branch = develop
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,8 @@ add_subdirectory(atmos_cubed_sphere)
 ### fv3atm
 ###############################################################################
 if(INLINE_POST)
+  set(BUILD_POSTEXEC OFF)
+  add_subdirectory(upp)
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -10,3 +10,6 @@
   path = ccpp/physics
   url = https://github.com/NCAR/ccpp-physics
   branch = main
+[submodule "upp"]
+  path = upp
+  url = https://github.com/DusanJovic-NOAA/UPP
```
WenMeng-NOAA: @DusanJovic-NOAA When UPP being submodule in the authoritative fv3atm, how will an UPP version be checked out?from the latest branch "develop" or tag "XXX"?

DusanJovic-NOAA: Submodules always point to specific commit, which can be a tag but it does not have to be. I guess most of the time, from the develop fv3atm branch we'll point to a commit in UPP develop branch. Production branch will probably point to a commit in UPP production branch.

WenMeng-NOAA: @DusanJovic-NOAA So in that case, I don't need create a tag from UPP repository site. We may check UPP version via hash number of upp directory on atmfv3 repository, right?

DusanJovic-NOAA: Correct. fv3atm 'knows' only about UPP url (optionally branch) and commit hash.

junwang-noaa (approved):

WenMeng-NOAA (approved):
Looks good to me. Thanks!

PR #422, opened 2021-11-10
Update several units in CCPP metadata following CCPP framework update

climbfuji:
## Description

With the metadata parser update in NCAR/ccpp-framework#415, a stricter checking of units will be enforced. This includes the replacement of "convenience units" such as log(Pa) with 1. The following changes are made in ccpp-physics and fv3atm:

- Replace units 'various' with 'mixed' as agreed upon on the ccpp-framework developer committee
- Update several invalid convenience units of non-physical quantities with `1`
- Correct the units of the relative humidity variables of the maximum hourly diagnostics from `%` to `frac` (because they are not percentages, but range from zero to one)
- Update submodule pointers for ccpp-framework and ccpp-physics

### Issue(s) addressed

n/a

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/907

## Dependencies

- waiting on https://github.com/NCAR/ccpp-framework/pull/415
- waiting on https://github.com/NCAR/ccpp-physics/pull/775

### DIFF
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -609,7 +609,7 @@
 [hprime]
   standard_name = statistical_measures_of_subgrid_orography_collection_array
   long_name = orographic metrics
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent,number_of_statistical_measures_of_subgrid_orography)
   type = real
   kind = kind_phys
@@ -5735,21 +5735,21 @@
 [tau_amf]
   standard_name = absolute_momentum_flux_due_to_nonorographic_gravity_wave_drag
   long_name = ngw_absolute_momentum_flux
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent)
   type = real
   kind = kind_phys
 [ozpl]
   standard_name = ozone_forcing
   long_name = ozone forcing data
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent,vertical_dimension_of_ozone_forcing_data,number_of_coefficients_in_ozone_forcing_data)
   type = real
   kind = kind_phys
 [h2opl]
   standard_name = stratospheric_water_vapor_forcing
   long_name = water forcing data
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent,vertical_dimension_of_h2o_forcing_data,number_of_coefficients_in_h2o_forcing_data)
   type = real
   kind = kind_phys
@@ -6340,7 +6340,7 @@
 [fluxr]
   standard_name = cumulative_radiation_diagnostic
   long_name = time-accumulated 2D radiation-related diagnostic fields
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent,number_of_diagnostics_variables_for_radiation)
   type = real
   kind = kind_phys
@@ -6921,7 +6921,7 @@
 [dtend]
   standard_name = cumulative_change_of_state_variables
   long_name = diagnostic tendencies for state variables
-  units = various
+  units = mixed
   dimensions = (horizontal_loop_extent,vertical_layer_dimension,cumulative_change_of_state_variables_outer_index_max)
   type = real
   kind = kind_phys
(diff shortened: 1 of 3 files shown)


ccpp/data/GFS_typedefs.meta
```diff
@@ -6957,14 +6957,14 @@
...
-  units = %
+  units = 1
   dimensions = (horizontal_loop_extent)
   type = real
   kind = kind_phys
 [rh02min]
   standard_name = minimum_relative_humidity_at_2m_over_maximum_hourly_time_interval
   long_name = minumum relative humidity at 2m over maximum hourly time interval
-  units = %
+  units = 1
```
junwang-noaa: May I ask what is the meaning of "units 1" and "units mixed"? Maybe you can see the units in NCEP grib2 product table as a reference? https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_table4-2-0-1.shtml

climbfuji: Relative humidity in the code is expressed between 0 and 1, i.e. as a fraction. If it was a real percentage, it would range from 0 to 100. See the actual code in ccpp-physics.

Note also we do have `percent` as a valid unit for such quantities. The `%` symbol is not acceptable as a unit in the CCPP, it causes all sorts of problem with parsing code (also because members of derived data types have a percent in the name, e.g. `foo%bar`). We could use `fraction` instead of `1` for relative humidity, but if so we need to go back to the CCPP framework developers and have them confirm.

Any variable that is a number without a physical unit can have unit `1`.

`mixed` is simply replacing `various` and is used in cases where a variable contains multiple units. For example, the grab-bag tracer array `qgrs` contains mass mixing ratios (`kg kg-1`), number concentrations `(kg-1)`, cloud amount (`fraction` or `1`, can't remember) etc. When the individual members of the `qgrs` array are passed, then the correct units can be used. But when the entire array is passed, we need to use `mixed` (was `various` until now).

Lastly, note that CCPP units do not follow the standard in https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/grib2_table4-2-0-1.shtml. If they do match, that is great, and we should strive for consistency where possible. But CCPP units follow CF conventions and `udunits`, with the extensions defined in https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst (as approved by the CCPP framework developer team).

climbfuji: Comparing the grib2 doc with what we have I see a nearly 100% match. The only difference is that we use `percent` instead of `%` for real percentages. Some units may differ from scheme to scheme (e.g. Thompson MP defines cloud effective radii in micron), but this is why we have automatic unit conversions - this way the host models can keep their units and still use the physics as-is.

junwang-noaa:
I think all the NCPE productions are using these tables (most fields are
from WMO except some local fields). For easy communication, can we use
"fraction" or "percent" instead of "1" in the host model then?

climbfuji:
We cannot use percent if the quantities range between 0 and 1. Percentages by definition range from 0 to 100. Using percent for "relative humidity" when it is bounded by zero and one is incorrect. We can use fraction, though.

climbfuji:
> We cannot use percent if the quantities range between 0 and 1. Percentages by definition range from 0 to 100. Using percent for "relative humidity" when it is bounded by zero and one is incorrect. We can use fraction, though.
> […](#)

If you are interested, here is the discussion about relative humidity in the ESCOMP CCPPStandardNames dictionary: https://github.com/ESCOMP/CCPPStandardNames/issues/26

climbfuji:
Here is the corresponding code in `maximum_hourly_diagnostics.F90` that shows that the variables in question in this PR are indeed bounded by 0 (or `RHmin = 1.0E-6`), and 1, not by 0 and 100:
```
...
          rh02=q2m(i)/QCQ
          IF (rh02 > 1.0) THEN
             rh02 = 1.0
          ENDIF
          IF (rh02 < RHmin) THEN !use smaller RH limit for stratosphere
             rh02 = RHmin
          ENDIF
          rh02max(i) = max(rh02max(i),rh02)
          rh02min(i) = min(rh02min(i),rh02)
```

climbfuji:
@junwang-noaa Please see this comment regarding `fraction` vs `1` - https://github.com/ESCOMP/CCPPStandardNames/issues/26#issuecomment-965887730

junwang-noaa (approved):

climbfuji:
I checked the hashes for ccpp-framework and ccpp-physics, they look correct to me. Ready to merge.

PR #434, opened 2021-12-01
Make surface emissivity over water a persistent variable

### LINKED ISSUE #914: all zero values in atmImp_Faxa_lwnet over ocean for cpld_control_p7 test
## Description
`atmImp_Faxa_lwnet` variable in the mediator history output has all zero over the ocean. This is probably passed to the other model components such as MOM6 that could cause an issue.

### To Reproduce:
What compilers/machines are you seeing this with? Intel+MPT on Cheyenne
Give explicit steps to reproduce the behavior.
1. checkout ufs-wearher model (head of develop)
2. add mediator history phase to the run sequence in nems.configure (`MED med_phases_history_write`)
3. run model
4. check `atmImp_Faxa_lwnet` in the mediator history file.

## Output
Screen Shot 2021-11-16 at 9 55 36 AM

climbfuji:
## Description

The surface emissivity over water is currently stored in the `GFS_interstitial_type` DDT, which is not persistent. To fix the issues described in https://github.com/ufs-community/ufs-weather-model/issues/914, it needs to be moved to a persistent data type. This PR moves it to the `GFS_sfcprop_type`, which is already home to the surface emissivity over land and ice.

Note that in the long run, in case restart reproducibility is broken, `Sfcprop%emis_wat` might need to be added to the restart files. Right now it doesn't matter, because the emissivity over water is just a constant value that gets set again and again each time step.

### Issue(s) addressed

Fixes https://github.com/ufs-community/ufs-weather-model/issues/914

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/921

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/787

### DIFF
--- a/io/module_wrt_grid_comp.F90
+++ b/io/module_wrt_grid_comp.F90
@@ -544,7 +544,7 @@
         wrt_int_state%lonlast  = lon(imo)
         do j=lbound(latPtr,2),ubound(latPtr,2)
           do i=lbound(lonPtr,1),ubound(lonPtr,1)
-            lonPtr(i,j) = 360.d0/real(imo,8) * real(i-1,8)
+            lonPtr(i,j) = lon(i)
             latPtr(i,j) = lat(j)
           enddo
         enddo
--- a/io/module_wrt_grid_comp.F90
+++ b/io/module_wrt_grid_comp.F90
@@ -537,8 +537,8 @@
         endif
         wrt_int_state%latstart = lat(1)
         wrt_int_state%latlast  = lat(jmo)
-        do j=1,imo
-          lon(j) = 360.d0/real(imo,8) *real(j-1,8)
+        do i=1,imo
+          lon(i) = 360.d0/real(imo,8) *real(i-1,8)
         enddo
         wrt_int_state%lonstart = lon(1)
         wrt_int_state%lonlast  = lon(imo)
--- a/io/module_wrt_grid_comp.F90
+++ b/io/module_wrt_grid_comp.F90
@@ -548,6 +548,28 @@
             latPtr(i,j) = lat(j)
           enddo
         enddo
+        wrt_int_state%lat_start = lbound(latPtr,2)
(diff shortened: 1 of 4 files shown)


climbfuji:
Adding @SMoorthi-emc only for his information later on, because he is on leave for the rest of the year.

io/module_wrt_grid_comp.F90
```diff
@@ -548,6 +548,28 @@ subroutine wrt_initialize(wrt_comp, imp_state_write, exp_state_write, clock, rc)
             latPtr(i,j) = lat(j)
```
DusanJovic-NOAA, line 548: Not strictly part of this PR, but I would suggest instead of recomputing lonPtr(i,j) (line 547) to use already computed values from array lon (line 541). BTW, for consistency, the index in that loop (540-542) should be `i`, not `j`.

climbfuji:
@DusanJovic-NOAA please check if https://github.com/NOAA-EMC/fv3atm/pull/434/commits/cb72fdae7eec367dbe836c4f5e03d149c6751f9c AND https://github.com/NOAA-EMC/fv3atm/pull/434/commits/91d48d7226ff3146e0b55ffcfe76dc3c7a7c8169 do what you suggested. Thanks!

DusanJovic-NOAA:
> @DusanJovic-NOAA please check if [cb72fda](https://github.com/NOAA-EMC/fv3atm/commit/cb72fdae7eec367dbe836c4f5e03d149c6751f9c) AND [91d48d7](https://github.com/NOAA-EMC/fv3atm/commit/91d48d7226ff3146e0b55ffcfe76dc3c7a7c8169) do what you suggested. Thanks!

Looks good to me.

DusanJovic-NOAA (approved):

junwang-noaa (approved):

climbfuji:
Hash for ccpp-physics is correct, please merge.

DeniseWorthen (approved):

PR #435, opened 2021-12-01
Merge NOAA-GSL gsl/develop into develop, update CCPP SDFs

climbfuji:
## Description

This PR updates many of the suite definition files using the RAP (previously: GSL) physics and add three additional variables in `GFS_typedefs.{F90,meta}` that are used with GF convection only. A few additional bug fixes are included in the CCPP metadata (`GFS_typedefs.{meta}`).

### Issue(s) addressed

See https://github.com/ufs-community/ufs-weather-model/issues/952

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/937

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/791


.gitmodules
```diff
@@ -8,8 +8,10 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/NCAR/ccpp-physics
-  branch = main
+  #url = https://github.com/NCAR/ccpp-physics
+  #branch = main
+  url = https://github.com/climbfuji/ccpp-physics
```
tanyasmirnova: @climbfuji Dom, Is this change pointing to your fork temporary?

tanyasmirnova (approved):

.gitmodules
climbfuji: Yep, just for the purpose of automated regression testing ... means that by adjusting `.gitmodules` all the way up to the top-level ufs-weather-model repository, one (person or bot) can check out the code recursively without making any changes.

climbfuji: This gets reverted once ccpp-physics is merged, and similar the change in the `.gitmodules` file in ufs-weather-model gets reverted when fv3atm is merged.

tanyasmirnova: Ok, thank you for explaining.

ccpp/data/GFS_typedefs.meta
```diff
@@ -5772,6 +5780,14 @@
   dimensions = (horizontal_loop_extent,vertical_layer_dimension,number_of_aerosol_tracers_MG)
   type = real
   kind = kind_phys
+[aod_gf]
+  standard_name = aod_gf_deep
```
grantfirl: @climbfuji reminder to change here too

ccpp/data/GFS_typedefs.F90
```diff
@@ -6027,7 +6032,11 @@ subroutine tbd_create (Tbd, IM, Model)
 
     if (Model%imfdeepcnv == Model%imfdeepcnv_gf) then
        allocate(Tbd%cactiv(IM))
+       allocate(Tbd%cactiv_m(IM))
+       allocate(Tbd%aod_gf(IM))
```
grantfirl, line 6056: @climbfuji @hannahcbarnes I see that this is only allocated for GF here, which strengthens the argument to qualify the standard name.

io/FV3GFS_io.F90
```diff
@@ -2787,8 +2791,8 @@ subroutine fv3gfs_diag_output(time, diag, atm_block, nx, ny, levs, ntcw, ntoz, &
          !---
          !--- skipping other 3D variables with the following else statement
          !---
-         if(mpp_pe()==mpp_root_pe())print *,'in,fv3gfs_io. 3D fields, idx=',idx,'varname=',trim(diag(idx)%name), &
-             'lcnvfac=',lcnvfac, 'levo=',levo,'nx=',nx,'ny=',ny
+!         if(mpp_pe()==mpp_root_pe())print *,'in,fv3gfs_io. 3D fields, idx=',idx,'varname=',trim(diag(idx)%name), &
```
climbfuji, line 2794: I commented this out to remove the clutter from stdout

io/FV3GFS_io.F90
```diff
@@ -783,7 +783,11 @@ subroutine sfc_prop_restart_read (Sfcprop, Atm_block, Model, fv_domain, warm_sta
       !--- allocate the various containers needed for restarts
       allocate(sfc_name2(nvar_s2m+nvar_s2o+nvar_s2mp+nvar_s2r))
       allocate(sfc_name3(0:nvar_s3+nvar_s3mp))
-      allocate(sfc_var2(nx,ny,nvar_s2m+nvar_s2o+nvar_s2mp+nvar_s2r),sfc_var3ice(nx,ny,Model%kice))
+      allocate(sfc_var2(nx,ny,nvar_s2m+nvar_s2o+nvar_s2mp+nvar_s2r))
+      ! Note that this may cause problems with RUC LSM for coldstart runs from GFS data
+      ! if the initial conditions do contain this variable, because Model%kice is 9 for
+      ! RUC LSM, but tiice in the initial conditions will only have two vertical layers
+      allocate(sfc_var3ice(nx,ny,Model%kice))
```
junwang-noaa, line 790: So the current RUC LSM test does not use GFS data and its cold start ICs has 9 levels?

DusanJovic-NOAA (approved):

io/FV3GFS_io.F90
climbfuji: No, this refers to the ice data. GFS v16 ICs do not contain the ice subsurface temperature
```
Sfcprop(nb)%tiice(ix,k) = sfc_var3ice(i,j,k)   !--- internal ice temp
```
in the coldstart data. Should, however, the tiice data from GFS sfc_sice physics be added to the coldstart ICs, then it would have only two layers, whereas RUC expects nine layers.

For restart/warmstart runs using the same physics, this is not a problem. tiice for RUC will be written with 9 layers and it will be read back in just fine.

junwang-noaa (approved):

climbfuji:
Hash for ccpp-physics is correct (4865cba), ready to merge

ccpp/suites/suite_FV3_RRFS_v1alpha.xml
```diff
@@ -1,84 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
JeffBeck-NOAA, line 1: While updating the SPP PR to the latest fv3atm code, I noticed this file was deleted (along with a few other suites, some used in the SRW App: FV3_GSD_v0, FV3_GSD_noah, and FV3_GSD_SAR). This particular suite is one of the two main SDFs in the SRW App release v1.0, and we should keep it for now.  FYI: @jwolff-ncar, @gsketefian, @mkavulich, @willmayfield, @BenjaminBlake-NOAA, @chan-hoo.

climbfuji: This was discussed with Curtis and others. The reasoning for removing it is that this suite is not going to be the operational RRFS suite in the future. Since code bases for the SRW App 1.0 and develop have diverged so far, it is not realistic to run the SRW App code with the head of the develop branches. If users have problems with the public release using RRFS v1alpha, you can't just point them to the develop branches and solve their questions there, because the code is vastly different. Therefore there is no harm in cleaning up old suite definition files.

It was decided to retain RRFS v1 beta for now. And many of the GSD_* suites have simply been renamed and rebased on the RAP suite.

JeffBeck-NOAA: @climbfuji, thanks for providing some background information.  Unfortunately, the DTC team that works on SRW App code management, development, and UFS-R2O RRFS testing wasn't aware of this discussion or decision.  

I'm certainly in favor of cleaning up old SDFs, but we just need to be careful.  In this case, since the v1beta SDF differs only in its use of MYNN SFC instead of GFS SFC, and since the authoritative code has diverged significantly from the release, the move to v1beta would make sense.  However, we've done a lot of RRFS benchmarking UFS-R2O work in the last few months with the v1alpha SDF in the latest SRW App, so it might be good for users to still have this SDF available in the latest authoritative code for testing/comparison to those results.  If that's not possible, then OK.  

Speaking more generally to the whole UFS community, it would be good to find a way to get all relevant voices included on important SRW App decisions like these in the future.

climbfuji: Are you pointing to the latest develop branches in the app? If not (e.g. if you are using your own forks), then you can simply retain the file for testing/benchmarking. It's a challenge to know for the non-regional people who needs to be consulted for which kind of changes. In this case, Curtis, Christina and several others (that I don't remember) were involved.

JeffBeck-NOAA:
Thanks, @climbfuji.  We are near the HEADs of the authoritative develop branches in the app (set in the Externals.cfg file).  We try to get them updated as frequently as possible, but there are times when they will lag behind for a while; just not enough staff/time to keep them up-to-date all the time.  And I agree, it's difficult to know who all to contact for changes when it's another App, component, etc.  I'll bring this up in our next DTC SRW App meeting to see what we can do.  Maybe a huge SRW App mailing list that's used sparingly is an option.

jwolff-ncar:
Agree with @JeffBeck-NOAA - Please keep if possible. Our work with this SDF is recent (completed in Oct) and still relevant. The DTC SRW App team (or some representative) should definitely be involved in these conversations in some way. We need to discuss further how to make sure this happens.

climbfuji:
If the suite is still required, how about @JeffBeck-NOAA adds it back in with the current fv3atm PR for the stochastic perturbations? Just note it in the PR title and description as an unrelated change that does not affect the model/regression testing.

Note that the RRFS v1 alpha suite won't be tested by the ufs-weather-model regression tests. It will be up to the SRW App developers to make sure that the suite continues to function, unfortunately.

JeffBeck-NOAA:
@climbfuji, @jwolff-ncar, I added the FV3_RRFS_v1alpha SDF to the [fv3atm PR](https://github.com/NOAA-EMC/fv3atm/pull/454) and modified the title/description accordingly.

PR #440, opened 2021-12-10
Code cleanup. Remove used code/variables. Fix minor inconsistencies.

### LINKED ISSUE #441: Code cleanup. Remove used code/variables. Fix minor inconsistencies.
There are a lot of unused code/variables, a lot of little inconsistencies. Time for some cleanup.
For exact code changes see PR #440.

DusanJovic-NOAA:
## Description

There are a lot of unused code/variables, a lot of little inconsistencies. Time for some cleanup.

* Remove ESMF Alarm and TimeInterval variables from module_fv3_config.
* Variables nfhmax, nfhmax_hf are unused outside InitializeAdvertise. Declared them as local variables.
* There is no need to keep duplicates of all time related variables in atm_int_state when we can easily access them from atm_int_state%Atm.
* Remove redundant call to fms_init
* Add few missing ESMF_LogFoundError checks in module_fcst_grid_comp.F90
* Delete time_utils.F90. Unused.
* print only actual errors to stderr, everything else to stdout
* Move realizeConnectedCplFields to module_cplfields from module_cap_cpl
* Declare Atmos as module variable, and remove atmos_internalstate_wrapper
* Move code from clock_cplIntval to InitializeAdvertise
* Removed INTERNAL_FILE_NML from atmos_model.F90

### Issue(s) addressed

- fixes #441

## Testing

## Dependencies

Do PRs in upstream repositories need to be merged first? No.


atmos_model.F90
```diff
@@ -491,23 +490,21 @@ subroutine atmos_model_init (Atmos, Time_init, Time, Time_step)
...
   integer :: isc, iec, jsc, jec
-  integer :: isd, ied, jsd, jed
-  integer :: blk, ibs, ibe, jbs, jbe
   real(kind=GFS_kind_phys) :: dt_phys
-  real, allocatable    :: q(:,:,:,:), p_half(:,:,:)
-  character(len=80)    :: control
-  character(len=64)    :: filename, filename2, pelist_name
-  character(len=132)   :: text
-  logical              :: p_hydro, hydro, fexist
+#ifndef INTERNAL_FILE_NML
```
junwang-noaa: I think dycore does not support this option any more. It has to use INTERNAL_FILE_NML even though the INTERNAL_FILE_NML are still in some of the code.

module_fcst_grid_comp.F90
```diff
@@ -839,78 +779,78 @@ subroutine fcst_run_phase_2(fcst_comp, importState, exportState,clock,rc)
...
 !-----------------------------------------------------------------------
 !
       tbeg1 = mpi_wtime()
-      rc    = esmf_success
+      rc    = ESMF_SUCCESS
 !
 !-----------------------------------------------------------------------
 !
-      call ESMF_GridCompGet(fcst_comp, name=compname, localpet=mype, rc=rc)
+      call ESMF_GridCompGet(fcst_comp, localpet=mype, rc=rc)
```
junwang-noaa, line 797: I think it's good to leave compname for debugging purpose

module_fcst_grid_comp.F90
```diff
@@ -737,16 +703,8 @@ subroutine fcst_initialize(fcst_comp, importState, exportState, clock, rc)
       call get_atmos_model_ungridded_dim(nlev=numLevels,         &
                                          nsoillev=numSoilLayers, &
                                          ntracers=numTracers)
-!
-!-----------------------------------------------------------------------
-!
-      IF(rc /= ESMF_SUCCESS) THEN
```
junwang-noaa, line 743: I think it's good to report if the fcst_initialize fails.

fv3_cap.F90
```diff
@@ -872,12 +889,11 @@ subroutine ModelAdvance(gcomp, rc)
...
     type(ESMF_Clock)            :: clock
     type(ESMF_Time)             :: currTime, startTime, stopTime
-    type(ESMF_TimeInterval)     :: timeStep
+    ! type(ESMF_TimeInterval)     :: timeStep
 
-    integer                     :: i, urc
     character(len=*),parameter  :: subname='(fv3_cap:ModelAdvance)'
     character(240)              :: msgString
-    character(240)              :: startTime_str, currTime_str, stopTime_str, timeStep_str
+    ! character(240)              :: startTime_str, currTime_str, stopTime_str, timeStep_str
```
junwang-noaa, line 896: I think in ModelAdvance, the integrate loop should be removed, we just need the code inside the loop.

fv3_cap.F90
```diff
@@ -406,7 +408,23 @@ subroutine InitializeAdvertise(gcomp, rc)
       line=__LINE__, file=__FILE__)) call ESMF_Finalize(endflag=ESMF_END_ABORT)
 
     ! Read in the FV3 coupling interval
-    call clock_cplIntval(gcomp, CF)
+    call ESMF_ConfigGetAttribute(config=CF, value=medAtmCouplingIntervalSec, &
```
junwang-noaa, line 411: I thought we don't need the code related to "atm_coupling_interval_sec" any more. Can you remove this section, my understanding is the code should still work.

DusanJovic-NOAA: We still do need this option. I tried to remove that entire section (former clock_cplIntval subroutine) but coupled tests fail. Removing this option will require some changes in how we handle clocks in the cap.

atmos_model.F90
DusanJovic-NOAA: Se are you saying we can remove this ifndef entirely? I still see a lot of similar ifdefs in ccpp and dycore

module_fcst_grid_comp.F90
DusanJovic-NOAA: This is not used consistently, for example in other routines like fcst_run_phase_1 and fcst_run_phase_1 this was commented out. Moreover if this routine fails, i.e. rc is not ESMF_SUCCESS at this moment calling code will report an error. In addition as soon as one of the preceding ESMF calls return we check rc and handle that error and return early, so this final check is probably never reached in case of an error.

fv3_cap.F90
DusanJovic-NOAA: Let me try that.

module_fcst_grid_comp.F90
DusanJovic-NOAA: It is unused. I can put it back, but my idea was to remove all unused variables so that we get clean compilation with `-Wall` compiler flag.

atmos_model.F90
junwang-noaa: Yes, GFDL will do some clean up. The code only works with INTERNAL_FILE_NML defined.

fv3_cap.F90
DusanJovic-NOAA: `cpld_debug_p7` test passed, but `control` failed.

junwang-noaa:
May I take a look at your run directory?

fv3_cap.F90
junwang-noaa: I see. So this is where the timestep is not available as Rocky mentioned. But the internal timestep is available at ModelAdvance, I am thinking of removing the atm_coupling_interval_sec and use the timestep at ModelAdvance to reset the clock, what do you think? Currently the atm_coupling_interval_sec has to be consistent with the inner time interval in the run sequence.

atmos_model.F90
DusanJovic-NOAA: Removed in d5409df

DusanJovic-NOAA:
> May I take a look at your run directory?

/scratch1/NCEPDEV/stmp2/Dusan.Jovic/FV3_RT/no_integration_loop_rt_130864/control

fv3_cap.F90
DusanJovic-NOAA: I agree that we should take a closer look at how clocks are defined and/or modified in the cap, and maybe see if we can eliminate the need for the cap to keep it's own local clock (`clock_fv3`). But I would like to keep this PR as it is and try to  address this issue in one of the future PRs. The current logic in the code related to clocks is not trivial, and it's not easy follow what's really going on.

fv3_cap.F90
junwang-noaa: Thanks for testing

junwang-noaa (approved):

PR #459, opened 2022-01-04
Feature/hwrf legacy

ChunxiZhang-NOAA:
## Description
Add two physics suites for the  HAFS. This is the second HAFS physics suite and it uses the GFDL surface layer scheme from the legacy of the HWRF suite and the Thompson microphysics scheme. 

## Testing
Created regression test. The test ran successfully.


climbfuji (approved):

junwang-noaa (approved):

binli2337:
@junwang-noaa Can you merge the PR? I don't have the permission.

junwang-noaa:
OK, will merge the code when all RT log files are committed in ufs-weather-model.

PR #480, opened 2022-02-10
Multiple output grids

DusanJovic-NOAA:
## Description

Update fv3 cap and write grid component to enable outputting multiple domains. This is done be creating an array of fcstGrids, and array of rout handles where each element of these arrays correspond to one atm domain. In the write grid component updates were made to allow grid spec parameters for each output grid to be specified separately. 

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
- fixes https://github.com/NOAA-EMC/fv3atm/issues/481

## Testing

How were these changes tested?  ufs regression test on hera
What compilers / HPCs was it tested with?  Intel.
Are the changes covered by regression tests? Yes.
Have the ufs-weather-model regression test been run? Yes. On what platform? Hera.  
- Will the code updates change regression test baseline? Yes. All multiple domain HAFS tests have new history outputs. And outputs of all global single domain tests will be chaged do to slight difference in how fcstGrid is now defined.
- Please commit the regression test log files in your ufs-weather-model branch

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)

Do PRs in upstream repositories need to be merged first?
If so add the "waiting for other repos" label and list the upstream PRs
- waiting on https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/176

PR #485, opened 2022-02-17
 Update inline_post_stub.F90 subroutine interfaces to match inline_post.F90

MicroTed:
## Description

Fixes compile error when INLINE_POST = OFF, which switches compilation from inline_post.F90 to inline_post_stub.F90 Updates no-op subroutines to match changes that were made in #480  

It seems the only way to trigger this (i.e., to have INLINE_POST = OFF) is to not have CMAKE_Platform set. Otherwise, all the platform defaults set it to ON.

## Testing

Tested on Mac and compiled on Jet

No changes in results expected.

### DIFF
--- a/io/inline_post_stub.F90
+++ b/io/inline_post_stub.F90
@@ -13,7 +13,7 @@ module inline_post
 
   contains
 
-  subroutine inline_post_run(wrt_int_state,mypei,mpicomp,lead_write,      &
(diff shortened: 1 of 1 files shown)


DusanJovic-NOAA:
Merged in https://github.com/NOAA-EMC/fv3atm/commit/8c582f7b1884b660e645e2e0756f457748ddbcec

PR #492, opened 2022-03-02
Add 'valid time' variable using ISO string format

### LINKED ISSUE #1052: Write out valid time as a string in history files
## Description
Currently, UFS write-grid component writes out valid time like so:
```
	double time(time) ;
		time:calendar = "JULIAN" ;
		time:calendar_type = "JULIAN" ;
		time:cartesian_axis = "T" ;
		time:long_name = "time" ;
		time:units = "hours since 2021-03-22 06:00:00" ;
```
This requires routines to compute the valid time using the value of time(t) and the unit string. These are in `w3emc` but is not a trivial thing to do without external libraries.

It would be nice to have something akin to the following added:
```
	char valid_time(time, nchars) ;
		valid_time:calendar = "JULIAN" ;
		valid_time:calendar_type = "JULIAN" ;
		valid_time:cartesian_axis = "T" ;
		valid_time:long_name = "valid time" ;
		valid_time:units = "ISO 8601 Date String" ;
```
So that one can easily grab: `2022-02-16T17:11:00Z`
This will be useful for a few things, most notably 1) checking in FV3-JEDI that the valid time from a cubed-sphere history file is at the expected time and 2) having an easy variable for plotting model output.

## Solution
Add a string representation of the model valid time to the history files.

## Alternatives
Add an integer representation of the model valid time to the history files.

## Related to
Adding cube-sphere history I/O to FV3-JEDI.

DusanJovic-NOAA:
## Description
Add 'valid time' variable using ISO string format to netcdf history files.
Use double precision variable to set value of 'time' attribute in wrt comp import state
Update ccpp/physics (setting surface-related interstitial variables for SCM prescribed-surface-flux mode)
Update inline_post_stub.F90 subroutine interfaces to match inline_post.F90

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/1052

## Testing

How were these changes tested?  Used one ufs regression test to verify that the actual data are identical.
What compilers / HPCs was it tested with?  Intel
Have the ufs-weather-model regression test been run? Not yet.
- Will the code updates change regression test baseline? Yes.
- If yes, why? New netcdf variable is added, so all netcdf history files will be different. 

## Dependencies

- waiting on https://github.com/NCAR/ccpp-physics/pull/870

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,8 +8,10 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/NCAR/ccpp-physics
-  branch = main
+  url = https://github.com/grantfirl/ccpp-physics
+  branch = sfc_flux_spec_update
+  #url = https://github.com/NCAR/ccpp-physics
+  #branch = main
 [submodule "upp"]
   path = upp
   url = https://github.com/NOAA-EMC/UPP
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,10 +8,8 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/grantfirl/ccpp-physics
-  branch = sfc_flux_spec_update
-  #url = https://github.com/NCAR/ccpp-physics
-  #branch = main
+  url = https://github.com/NCAR/ccpp-physics
+  branch = main
 [submodule "upp"]
   path = upp
   url = https://github.com/NOAA-EMC/UPP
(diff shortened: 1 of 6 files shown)


CoryMartin-NOAA (approved):
Thanks for adding this!

aerorahul (approved):
LGTM!

junwang-noaa (approved):

junwang-noaa:
@DusanJovic-NOAA CCPP physics PR was merged.

PR #55, opened 2020-02-06
ufs_public_release: remove libxml2

climbfuji:
Remove reference to libxml2, this not required for the CCPP static build. Avoids that users need to install libxml2 on their systems for something that doesn't get used at all.

Also updates the submodule pointer for atmos_cubed_sphere following the merge of the fv3dycore scientific documentation, and contains the update of the branch names for ccpp/framework and ccpp/physics from an earlier commit (from ufs_public_release to release/public-v4).

Associated PRs:

https://github.com/NCAR/ccpp-framework/pull/261
https://github.com/NOAA-EMC/fv3atm/pull/55
https://github.com/ufs-community/ufs-weather-model/pull/46

For regression testing, see https://github.com/ufs-community/ufs-weather-model/pull/46.

PR #513, opened 2022-03-25
Feature/lndp noahmp2 : add land perturbation scheme for Noah-MP fractional veg 

ClaraDraper-NOAA:
## Description

Updates to allow land perturbation scheme to be applied for Noah-MP. Currently, for Noah-MP model parameters can perturb only vegetation fraction. 

Main code changes: 
-added lsm_noahmp model option to lndp_type==2 land perturbation scheme  (fv3-atm, stochastic_physics)
-updated comments
-cleaned up the namelists to make the different options clearer. Removed misleading lndp_each_step variable from gfs_physics_nml, and replaced it with "lndp_model_type" in the nam_sfcperts namelist for the different forecast types (cycling DA, short forecasts, perturbing only initial conditions)   (fv3-atm, stochastic_physics, ufs-weather-model)
NOTE: see note below, re: specification of smc perturbation for RAP/HRRR/etc (lndp_model_type==2)
-revised code in stochastic_physics_wrapper to only allocate arrays that will be used when passed into lndp_apply_perts (fv3-atm) 
-deleted unused albedo arrays in stochastic_physics_wrapper and lndp_apply_perts (fv3-atm, stochastic_physics) 
-fixed bug causing precision errors when calculating the soil moisture ice content (often ~0.0; stochastic_physics) 
-fixed bug in which lndp_apply_perts wasn't called for nscyc = 0. (fv3-atm) 
-increased max_nvar_lndp at Jeff Ator's request (fv3-atm) 
-added  a new test for the Noah-MP and lndp==2 combination (ufs-weather-model) 

### Issue(s) addressed

Addresses issue #55 in stochastic_physics. 

## Testing

All tests passed on here, except the lndp ones (see below). 

Changes to tests: 
-added new control_p8_lndp to test Noah-MP land perts 
-updated previous lndp tests with new results, now that  precision errors when calculating the soil moisture ice content is fixed 
-also, for rap_lndp_debug the perturbation to smc at the initial conditions (applied only once) was previously specified as a rate, then divided by the time step before being applied. This makes the magnitude of perturbation dependent on the model time step. This has been fixed with the introduction of the lndp_model_type==2 option for regional sysyems, so that the perturbation is now specified as an absolute amount (changed 0.2 to 0.017 in the test). The previously used smc pert needs to be multiplied by timestep/3600 to get the same results with the new code.

New tests are here on hera: 
/scratch2/BMC/gsienkf/Clara.Draper/stmp4/Clara.Draper/FV3_RT/REGRESSION_TEST_INTEL

## Dependencies

Three linked PRS: 
stochastic_physics ( Feature/lndp noahmp2 : add land perturbation scheme for Noah-MP fractional veg https://github.com/noaa-psd/stochastic_physics/pull/56)
ufs-weather-model (add land perturbation scheme for Noah-MP fractional veg [#1143](https://github.com/ufs-community/ufs-weather-model/pull/1143 )
fv3-atm ( Feature/lndp noahmp2 : add land perturbation scheme for Noah-MP fractional veg https://github.com/NOAA-EMC/fv3atm/pull/513)


ccpp/data/GFS_typedefs.F90
```diff
@@ -1161,27 +1161,25 @@ module GFS_typedefs
...
     integer              :: spp_pbl
     integer              :: spp_sfc
     integer              :: spp_mp
     integer              :: spp_rad
     integer              :: spp_gwd
     integer              :: n_var_spp
-    character(len=3)    , pointer :: spp_var_list(:)  ! dimension here must match n_var_spp in stochy_nml_def
-    real(kind=kind_phys), pointer :: spp_prt_list(:)  ! dimension here must match n_var_spp in stochy_nml_def 
-    real(kind=kind_phys), pointer :: spp_stddev_cutoff(:)  ! dimension here must match n_var_spp in stochy_nml_def 
+    character(len=3)    , pointer :: spp_var_list(:)
```
junwang-noaa, line 1180: Is it still true that the spp_var_list needs to have dimension n_var_spp?

ClaraDraper-NOAA: Yes, but I don't think the comment is required anymore. It looks to me like these lines have been copied from my old lndp_var_list and lndp_prt_list code and then adapted to spp. Originally lndp_var_list and lndp_prt_list were specified here as arrays and not pointers. The same array is also specified in stochastic_physics, so the dimension of the array here had to match that specified in stochastic physics - hence I added the comment. Now that lndp_var/prt_list and spp_var/prt_list are pointers I don't think we need the comment anymore.

junwang-noaa (approved):

PR #523, opened 2022-04-20
Add prognostic cumulus closure description in saSAS

lisa-bengtsson:
## Description
This PR introduces an optional closure assumption to the Arakawa-Schubert cloud work function in saSAS cumulus convection, as well as the GFS shallow convection scheme, in where the mass-flux at cloud base is described as the product between updraft area fraction and updraft velocity. The updraft area fraction is given by a prognostic moisture budget equation.

The update is not aimed for UFS coupled prototype 8, but is an option to be explored beyond P8.

In addition to updates in the ccpp/physics submodule, new field_table, and namelist parameters are added in ufs-weather-model.

The PR is also making a small change in the stochastic_physics_driver.F90 to allow the cellular automata to work on non-uniform blocksizes. Previously the code aborted if not all blocks in the domain had the same size, however, the code works the way it's been formulated, so I simply removed the call to mpp_error in case on non-uniform blocksizes. 

### Issue(s) addressed
Issue: https://github.com/NCAR/ccpp-physics/issues/890
https://github.com/ufs-community/ufs-weather-model/issues/1193

## Testing
This PR does not change the answers in the control, but a new test has been added where the scheme is activated using progsigma = True.

The regression tests have been tested on Hera. 

## Dependencies

Dependency: 
https://github.com/NCAR/ccpp-physics/pull/903
https://github.com/ufs-community/ufs-weather-model/pull/1187


ccpp/data/CCPP_typedefs.F90
```diff
@@ -1141,8 +1141,12 @@ subroutine gfs_interstitial_setup_tracers(Interstitial, Model)
...
 
     if (Model%ntiw > 0) then
-      if (Model%ntclamt > 0) then
+      if (Model%ntclamt > 0 .and. Model%ntsigma == 0) then
         Interstitial%nn = Model%ntrac - 2
+      elseif (Model%ntclamt == 0 .and. Model%ntsigma > 0) then
+        Interstitial%nn = Model%ntrac - 2
+      elseif  (Model%ntclamt > 0 .and. Model%ntsigma > 0) then
+        Interstitial%nn = Model%ntrac - 3
       else
```
lisa-bengtsson, line 1150: @JongilHan66 @ChunxiZhang-NOAA this statement had me a bit puzzled, I'm not sure this is the right solution. Essentially I do not want to vertically diffuse updraft area fraction in PBL or convection, and I am aiming at treating ntsigma as ntclamt in this regard. So if either is present it should be reduced by 2, if both are present it should be reduced by 3. Maybe there is a better way?

ChunxiZhang-NOAA: @lisa-bengtsson Do you want to advect updraft area fraction in the dynamics?

lisa-bengtsson: @ChunxiZhang-NOAA yes, I do. And I'd made sure that is done correctly. I'm speaking of the handling of the vertical transport in physics. This hard coded segment is perhaps a bit dangerous as there are many combinations of tracers in the field tables. Ideally perhaps a flag could go with each tracer whether to diffuse or advect or both, but right now I'm just adding to this existing code.

ChunxiZhang-NOAA: @lisa-bengtsson I was thinking if updraft area fraction was not advect in dynamics, it didn't need to be defined as tracer, which could make things simple. Tracers are treated in a complicated way in CCPP.  Maybe we can find a better way in the future. Back to the current code, I believe the index number for missing one is -99.  It would be safer to use <=0.

lisa-bengtsson: @ChunxiZhang-NOAA ok, yes that is a good idea, I will change to <=. 

Yes the discussion on tracer treatment within the physics is beyond the scope of this PR, but I hope we can avoid these type of if-statements in the future.  Thanks. I will address this and your other comments today. Thanks for having a careful look.

JongilHan66: @lisa-bengtsson I don't get it on why you advect the sub-grid updraft fraction horizontally and vertically by the mean wind. Does it physically make sense? As I remember, the original equation for the prognostic updraft equation does not have advection term.

lisa-bengtsson: @JongilHan66 in Gerard et al. 2009 the area fraction is advected, also in the 3MT scheme. There is no advection term, but area fraction from the previous time-step is advected before going into termA. This is also what is done in our paper.

JongilHan66: @lisa-bengtsson it's hard to understand, considering the cumulus updraft is a sub-grid scale process. How can the sub-grid fraction go across the grid box? Also, the source term is moisture convergence, which is a result of moisture advection.

lisa-bengtsson: @JongilHan66 I suppose it would be a similar argument for advecting turbulent kinetic energy? It is a sub-grid scale process with local buoyancy/shear production and dissipation. If a larger area fraction represents "stronger" or "more organized" convection, I think it would make sense to advect this with the flow as e.g. meso-scale convective systems could be moved with the flow.

.gitmodules
```diff
@@ -1,16 +1,15 @@
...
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/NCAR/ccpp-physics
+  url = https://github.com/lisa-bengtsson/ccpp-physics
   branch = main
 [submodule "upp"]
   path = upp
   url = https://github.com/NOAA-EMC/UPP
   branch = develop
+[submodule "atmos_cubed_sphere"]
```
grantfirl: Unless someone asked to change the order of the submodules, I'd revert this.

.gitmodules
```diff
@@ -1,16 +1,15 @@
...
-  branch = dev/emc
 [submodule "ccpp/framework"]
   path = ccpp/framework
   url = https://github.com/NCAR/ccpp-framework
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/NCAR/ccpp-physics
+  url = https://github.com/lisa-bengtsson/ccpp-physics
   branch = main
```
grantfirl: You need to change the name of the branch to your ccpp-physics PR branch, prog_closure, otherwise this is telling git to grab the main branch on your fork.

junwang-noaa (approved):

grantfirl:
@lisa-bengtsson https://github.com/NCAR/ccpp-physics/pull/903 has been merged. Please pull down the latest ccpp-physics main branch, git add that latest branch, and revert your changes to .gitmodules in preparation for merging this PR.

lisa-bengtsson:
I have added the https://github.com/NCAR/ccpp-physics/pull/903 and reverted .gitmodules, I still see the changes in ccpp/physics here (in fv3atm), how can I see that it is pointing to the correct ccpp/physics?

grantfirl:
> I have added the [NCAR/ccpp-physics#903](https://github.com/NCAR/ccpp-physics/pull/903) and reverted .gitmodules, I still see the changes in ccpp/physics here (in fv3atm), how can I see that it is pointing to the correct ccpp/physics?

@lisa-bengtsson You *should* see that there are changes in ccpp/physics because you've updated the ccpp-physics submodule. The latest ccpp-physics commit hash is https://github.com/NCAR/ccpp-physics/commit/77bcfb1b39f7e737ccffee842031dae228240227 and it looks like your branch is correctly pointing to it, so we're all good.

You can check this on GitHub by looking at the 'files changed' in the PR, clicking on the ccpp/physics and then the number of changed files (19 files). It'll show which commit has fv3atm is pointing to. You can also use git in the command line to do this. Assuming that your latest fv3atm branch is checked out, if you do a `git submodule update --init --recursive` when in the FV3 directory, it will checkout the commits that your fv3atm branch is pointing to. Then, cd into ccpp/physics and run `git log` to see the commits.

I'm sure that there are other ways as well.

Edit: I just googled, and you can do `git submodule status` in your fv3atm branch to see which commits of the submodules that you're pointing to.

lisa-bengtsson:
> > I have added the [NCAR/ccpp-physics#903](https://github.com/NCAR/ccpp-physics/pull/903) and reverted .gitmodules, I still see the changes in ccpp/physics here (in fv3atm), how can I see that it is pointing to the correct ccpp/physics?
> 
> @lisa-bengtsson You _should_ see that there are changes in ccpp/physics because you've updated the ccpp-physics submodule. The latest ccpp-physics commit hash is [NCAR/ccpp-physics@77bcfb1](https://github.com/NCAR/ccpp-physics/commit/77bcfb1b39f7e737ccffee842031dae228240227) and it looks like your branch is correctly pointing to it, so we're all good.
> 
> You can check this on GitHub by looking at the 'files changed' in the PR, clicking on the ccpp/physics and then the number of changed files (19 files). It'll show which commit has fv3atm is pointing to. You can also use git in the command line to do this. Assuming that your latest fv3atm branch is checked out, if you do a `git submodule update --init --recursive` when in the FV3 directory, it will checkout the commits that your fv3atm branch is pointing to. Then, cd into ccpp/physics and run `git log` to see the commits.
> 
> I'm sure that there are other ways as well.
> 
> Edit: I just googled, and you can do `git submodule status` in your fv3atm branch to see which commits of the submodules that you're pointing to.

Great, thank you Grant. Then I think we are good to go with this branch.

DusanJovic-NOAA (approved):

PR #527, opened 2022-04-27
Change RRTMGP to RRTMG in two P8 SDFs

ChunxiZhang-NOAA:
## Description

Due to the low computational efficiency of the current RRTMGP scheme, the RRTMGP scheme in suite_FV3_GFS_v17_p8 and suite_FV3_GFS_v17_coupled_p8 were reverted to RRTMG.

### Issue(s) addressed
Low computational efficiency of the current RRTMGP scheme is avoided.

## Testing
Regression tests were conducted.

## Dependencies
NO


ccpp/suites/suite_FV3_HAFS_v0_thompson_tedmf_gfdlsf.xml
```diff
@@ -75,11 +75,7 @@
       <scheme>cnvc90</scheme>
       <scheme>GFS_MP_generic_pre</scheme>
       <scheme>mp_thompson_pre</scheme>
-    </subcycle>
-    <subcycle loop="2">
```
junwang-noaa, line 79: Does this mean the hafs v0 will not use subcycle 2 for Thompson MP?

yangfanglin:
Jun,  for Thompson MP, there are two options, one is subccyling and the other is inner-looping (controlled by a namelist option), to improve its computational instability.  Their effects are equivalent although not identical.   I'd recommend we use inner-looping for all schemes for consistency and also to avoid changing the SDF files.

junwang-noaa (approved):

DusanJovic-NOAA (approved):

PR #531, opened 2022-05-06
modify fhzero logic in emptying buckets

pjpegion:
## Description

The current logic in CCPP_driver does not allow for buckets to be emptied every time-step.
This one-line modification fixes that.

### Issue(s) addressed

- fixes noaa-emc/fv3atm/issues/#530

## Testing

running 24-hour forecasts, saving every time-step. Time-averaged fields no longer accumulate with time.

Hera/intel
Are the changes covered by regression tests? 
No, and I am not sure if we want an RT with this.

Limited set have been run on hera.
-NO

## Dependencies

None


ccpp/driver/CCPP_driver.F90
```diff
@@ -149,7 +149,7 @@ subroutine CCPP_step (step, nblks, ierr)
       endif
 
       !--- determine if physics diagnostics buckets need to be cleared
-      if (mod(GFS_control%kdt,GFS_control%nszero) == 1) then
+      if ((mod(GFS_control%kdt-1,GFS_control%nszero)+1) == 1) then
```
climbfuji: How is that different from
```
if (mod(GFS_control%kdt,GFS_control%nszero) == 0) then
```
?

pjpegion:
I assume going from mod 1 to mod 0 would change the baseline, since the buckets would be emptied 1 timestep earlier

climbfuji:
> I assume going from mod 1 to mod 0 would change the baseline, since the buckets would be emptied 1 timestep earlier
> […](#)

Yes, of course, you are right.

climbfuji (approved):

ccpp/driver/CCPP_driver.F90
DusanJovic-NOAA: How about:

`if (mod(GFS_control%kdt-1,GFS_control%nszero) == 0) then`

pjpegion: @DusanJovic-NOAA that makes sense.  I will test to confirm it gives the answer I expect.

pjpegion: @DusanJovic-NOAA's recommendation works.  Thanks!

DusanJovic-NOAA (approved):

PR #538, opened 2022-05-20
thompson_fluxcalc fix

bbakernoaa:
When switching to to Thompson in P8b with the coupled aerosols the values were not calculated by default and therefore zeros were passed to the GOCART model through the NUOPC layer causing no large scale wet deposition. This PR adds back in the capability that was added to the GFDL MP for the coupled GOCART model for the calculation of the large scale wet deposition. Here we use the instantaneous tendency to calculate the flux. This has no impact on the meteorology and only adds in the needed variables for the GOCART model.

See https://github.com/NCAR/ccpp-physics/pull/926 for more details


junwang-noaa:
@bbakernoaa The changes are in Chunxi's FV3ATM PR #525, right?  Thanks

bbakernoaa:
@junwang-noaa Yes thats correct.

junwang-noaa:
We will commit PR#525. You can close this PR. Thanks

bbakernoaa:
@junwang-noaa Thanks.  PR#525 still needs to pass the ORT correct?

PR #569, opened 2022-08-02
inline post updates for P8 and ESMF -managed threading

junwang-noaa:
## Description

Changes in the this PR includes:
1) With the ESMF -managed threading, the number of mpi tasks for UPP needs to be consistent with the traditional threading. To do that the write tasks per group used in UPP needs to set correctly considering the ESMF threading.  
2) Inline post interface update to resolve issues with missing post fields on pv and sigma levels when running in 32BIT=NO.
3) Update upp to the latest develop branch to fix the pressure level TCDC fields when the cldfra/cld_amt is not in the history output.
 
### Issue(s) addressed

- Fixes issue https://github.com/ufs-community/ufs-weather-model/issues/1304
- Fixes issue https://github.com/ufs-community/ufs-weather-model/issues/1320

## Testing

How were these changes tested? Tested on hera 

## Dependencies

https://github.com/ufs-community/ufs-weather-model/pull/1305

### DIFF
--- a/io/post_fv3.F90
+++ b/io/post_fv3.F90
@@ -74,7 +74,7 @@ module post_fv3
       integer,allocatable  :: istagrp(:),iendgrp(:)
       integer,save         :: kpo,kth,kpv
       logical,save         :: log_postalct=.false.
-      real,dimension(komax),save :: po, th, pv
+      real(4),dimension(komax),save :: po, th, pv
       logical        :: Log_runpost
       character(255) :: post_fname*255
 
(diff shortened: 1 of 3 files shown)


WenMeng-NOAA (approved):
The UPP part looks good to me.

WenMeng-NOAA:
@DusanJovic-NOAA 's PR #570 has a decency on the UPP PR #538. If this PR is merged first, no upp revision update is needed in that PR.

SMoorthi-emc (approved):

DusanJovic-NOAA (approved):

io/post_fv3.F90
```diff
@@ -2996,29 +2996,29 @@ subroutine set_postvars_fv3(wrt_int_state,mpicomp,setvar_atmfile,   &
 
             ! model level cloud fraction
             if(imp_physics == 11) then !GFDL MP
-            if(trim(fieldname)=='cld_amt') then
```
ChunxiZhang-NOAA, line 2999: Is it possible to replace the conditions 
       if(imp_physics == 11) then !GFDL MP
            if(trim(fieldname)=='cld_amt') then
            end if
       else
            if(trim(fieldname)=='cldfra') then
            end if
       end if
        with
       if(trim(fieldname)=='cld_amt' .or trim(fieldname)=='cldfra') then
       end if

junwang-noaa: I think's it's good not doing that as for GFDLMP you can output both "cld_amt" and "cldfra" in the diag_table, and the cfr will end up using cldfra which is not recommended for GFDLMP.

theurich (approved):
I can only really comment on the changes related to ESMF-managed threading in fv3_cap.F90. Looks all good to me.

io/post_fv3.F90
ChunxiZhang-NOAA: Ok, I see.

BrianCurtis-NOAA (approved):

PR #606, opened 2022-12-01
Add CLM Lake Model and update Flake

SamuelTrahanNOAA:
## Description

This PR adds the CLM Lake Model, updates the Flake This PR adds the CLM Lake Model, updates the Flake model, and adds an experimental feature to run both NSST and a lake model on lake points. The FV3_HRRR suite uses clm_lake by default, and the long-forgotten control_flake test now exists. Any test with LKM=1 that validates a restart file will change results.

### Issue(s) addressed

https://github.com/ufs-community/ufs-weather-model/issues/1510

## Testing

How were these changes tested?  *Manual review of output, and running regression tests.*
What compilers / HPCs was it tested with?   *hera.intel and hera.gnu*
Are the changes covered by regression tests?  *several tests are updated and a few new ones are added*
Have the ufs-weather-model regression test been run? On what platform?  *hera.intel and hera.gnu*
- Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below. *Any tests with LKM=1*

## Dependencies

https://github.com/ufs-community/ccpp-physics/pull/26 to add the CLM Lake Model


ccpp/driver/GFS_restart.F90
```diff
@@ -108,14 +115,22 @@ subroutine GFS_restart_populate (Restart, Model, Statein, Stateout, Sfcprop, &
     ! NoahMP
     if (Model%lsm == Model%lsm_noahmp) then
       Restart%num2d = Restart%num2d + 10
+      surface_layer_saves_rainprev = .true.
     endif
     ! RUC 
     if (Model%lsm == Model%lsm_ruc) then
       Restart%num2d = Restart%num2d + 5
+      surface_layer_saves_rainprev = .true.
```
tanyasmirnova, line 136: @SamuelTrahanNOAA Does it mean that without setting surface_layer_saves_rainprev =.true. here, there is a problem with the precipitation in the restart when MYNN surface layer is not used?

SamuelTrahanNOAA: The CLM Lake Model needs the two rain prev variables. I don't want to store them more than once though, so there's an "if" statement to store them only when the surface layer doesn't.

tanyasmirnova (approved):

io/FV3GFS_io.F90
```diff
@@ -2127,9 +2249,34 @@ subroutine sfc_prop_restart_write (Sfcprop, Atm_block, Model, fv_domain, timesta
...
+       sfc_name2(nvar2me+8)  = 't_bot1'
+       sfc_name2(nvar2me+9)  = 't_bot2'
+       sfc_name2(nvar2me+10)  = 'c_t'
+     else if(Model%iopt_lake==Model%iopt_lake_clm) then
+       ! Tell clm_lake to register all of its fields
+       call clm_lake%register_fields(Sfc_restart)
+     endif
+   endif
+   
    !--- register the 2D fields
```
HelinWei-NOAA: Add the following section after line 2321. Chunxi has successfully run control_flake_restart with this revision.
!Flake
   if (Model%lkm > 0 .and. Model%iopt_lake==Model%iopt_lake_flake) then
        nvar_before_lake=nvar2m+nvar2o+nvar2r+nvar2mp
      mand = .false.
        do num = nvar_before_lake+1,nvar_before_lake+nvar2l
         var2_p => sfc_var2(:,:,num)
         call register_restart_field(Sfc_restart, sfc_name2(num),var2_p,dimensions=(/'xaxis_1', 'yaxis_1', 'Time   '/),&
                                     &is_optional=.not.mand)
       enddo
   endif

SamuelTrahanNOAA: Although the flake restart does not crash, it also does not reproduce the baseline. The 2threads, decomp, and restart tests all do not match the baseline. This test was with hera.gnu. For your convenience, I've added a `tests/flake.conf` with all of the flake tests. You'll need to generate baselines (rt.sh -c) and compare against them (rt.sh -m).

HelinWei-NOAA: Yes. The results are quite different. I am debugging now. Did you run any clm lake restart case? Does it reproduce?

SamuelTrahanNOAA: With the FV3_HRRR suite on C96, all test combinations pass. With the rrfs_conus13km cases, the restart matches, but 2threads doesn't. (The decomp is already known to not match.)

I haven't tested clm lake with GFS v16 in a while.

SamuelTrahanNOAA: I'm putting together an rrfs.conf file with all of the relevant tests so you can use the target model, instead of gfs v16. It'll have flake, clm lake, and no lake, with debug & non-debug, and the other usual test variants. It'll take a few hours for me to test it.

HelinWei-NOAA:
Sam,Is it possible to have any clm_lake restart tests? If they can go through then we know the flake_restart problem is not because of the infrastructure. We can focus on flake model to sort it out. Thanks.HelinSent from my iPhoneOn Mar 9, 2023, at 10:49 AM, Samuel Trahan (NOAA contractor) ***@***.***> wrote:﻿
@SamuelTrahanNOAA commented on this pull request.



In io/FV3GFS_io.F90:
> +       sfc_name2(nvar2me+1) = 'T_snow'
+       sfc_name2(nvar2me+2) = 'T_ice'
+       sfc_name2(nvar2me+3)  = 'h_ML'
+       sfc_name2(nvar2me+4)  = 't_ML'
+       sfc_name2(nvar2me+5)  = 't_mnw'
+       sfc_name2(nvar2me+6)  = 'h_talb'
+       sfc_name2(nvar2me+7)  = 't_talb'
+       sfc_name2(nvar2me+8)  = 't_bot1'
+       sfc_name2(nvar2me+9)  = 't_bot2'
+       sfc_name2(nvar2me+10)  = 'c_t'
+     else if(Model%iopt_lake==Model%iopt_lake_clm) then
+       ! Tell clm_lake to register all of its fields
+       call clm_lake%register_fields(Sfc_restart)
+     endif
+   endif
+   
    !--- register the 2D fields

I'm putting together an rrfs.conf file with all of the relevant tests so you can use the target model, instead of gfs v16. It'll have flake, clm lake, and no lake, with debug & non-debug, and the other usual test variants. It'll take a few hours for me to test it.

SamuelTrahanNOAA:
The rrfs conus13km tests will include restart tests for flake, clm lake, and no lake. I'm working on this now.

SamuelTrahanNOAA:
@HelinWei-NOAA - Sorry for the delay. I have two sets of tests in these files:

1. hrrr_clm_lake.conf - FV3_HRRR suite with clm_lake at C96 using the lake data you provided. All tests pass with both gnu and intel compilers on hera.
2. rrfs.conf - Three configurations. All are 13km domain, warm start, with *no lake data*. The three sets are: flake, clm_lake, and no lake (clm_lake with LKM=0). Results are the same for both compilers (gnu & intel) on hera:
   - restart - runs but does not reproduce control
   - debug - runs
   - control - runs
   - decomp & debug_decomp - runs, but does not reproduce control; this is expected for the RRFS.
   - 2threads & debug_2threads - runs, but:
      - flake: debug_2threads does *not* match debug, but 2threads matches control
      - clm_lake: 2threads does *not* match control, but debug_2threads matches debug
      - no_lake: both match control

I did not try a debug_restart.

Note that the second set, rrfs.conf, has no lake data. That means the lake model cannot be the cause of the problem.

HelinWei-NOAA:
@SamuelTrahanNOAA so even the lake model is not turned on, the restart run can't reproduce? If this is the case, then it is still an infrastructure issue of lake coupling. For example, lake/ocean flag changes during the run.


> @HelinWei-NOAA - Sorry for the delay. I have two sets of tests in these files:
> 
> 1. hrrr_clm_lake.conf - FV3_HRRR suite with clm_lake at C96 using the lake data you provided. All tests pass with both gnu and intel compilers on hera.
> 2. rrfs.conf - Three configurations. All are 13km domain, warm start, with _no lake data_. The three sets are: flake, clm_lake, and no lake (clm_lake with LKM=0). Results are the same for both compilers (gnu & intel) on hera:
>    
>    * restart - runs but does not reproduce control
>    * debug - runs
>    * control - runs
>    * decomp & debug_decomp - runs, but does not reproduce control; this is expected for the RRFS.
>    * 2threads & debug_2threads - runs, but:
>      
>      * flake: debug_2threads does _not_ match debug, but 2threads matches control
>      * clm_lake: 2threads does _not_ match control, but debug_2threads matches debug
>      * no_lake: both match control
> 
> I did not try a debug_restart.
> 
> Note that the second set, rrfs.conf, has no lake data. That means the lake model cannot be the cause of the problem.

SamuelTrahanNOAA:
> so even the lake model is not turned on, the restart run can't reproduce? If this is the case, then it is still an infrastructure issue of lake coupling. For example, lake/ocean flag changes during the run.

A lake coupling infrastructure issue is the most likely explanation, but I haven't found any errors in the lake coupling code. I need you to look through and search for problems.

There is a more ominous possibility. The lake coupling framework could be 100% correct, but something else in the code is buggy. The lake coupling changes just happen to trigger that bug. UPP had an out-of-bounds access bug that was triggered by the clm lake changes (actually, it has many, but I only had to fix one).

HelinWei-NOAA:
dry/wet/icy should be defined in GFS_surface_composites_pre.F90 and shouldn't be redefined or changed elsewhere. After removing the redefining lake as wet(lake could be icy, so this will cause some problem) in GFS_surface_composites_post.F90, GFS_surface_composites_inter.F90, and sfc_diff.f,
control_restart_p8 reproduces control_p8 successfully.

SamuelTrahanNOAA:
The control_restart_p8 isn't the test I am debugging because that test *already* reproduces its control. What isn't working is the lake cases. You will find the flake cases in tests/flake.conf. The control and debug work, but no variants match their controls. Unfortunately, the changes you gave me don't fix any of them.

HelinWei-NOAA:
control_restart_p8 didn't reproduce when I tested. flake model itself still has some issues. I am working on it. I want to know if the clm_lake restart can reproduce after my changes.

SamuelTrahanNOAA:
None of the rrfs conus13km tests were able to reproduce, but that may be due to other issues.

All of the C96 FV3_HRRR suite tests use the clm_lake and they pass. That includes restart, decomp, 2threads, debug, and a mix of 32 and 64 bit physics and dynamics. What doesn't work is the RRFS 2threads tests, which did work before the lake changes.

SamuelTrahanNOAA:
@HelinWei-NOAA @tanyasmirnova - The threading issues were caused by an error in GFS_restart.F90. I think everything works now, so I've taken the PRs out of draft status.

I'm updating the subcomponents, and cleaning up the new test scripts. I'll do another round of testing after that.

tanyasmirnova:
@SamuelTrahanNOAA Great job, Sam! thank you very much for finding the problem with threading!

tanyasmirnova (approved):

SamuelTrahanNOAA:
This is scheduled for commit today, but it hasn't been reviewed.

ccpp/data/GFS_typedefs.F90
```diff
@@ -2211,6 +2296,25 @@ subroutine sfcprop_create (Sfcprop, IM, Model)
     Sfcprop%soiltype_frac = clear_val
     Sfcprop%lakefrac  = clear_val
     Sfcprop%lakedepth = clear_val
+
+    Sfcprop%use_lake_model = zero
+    if(Model%lkm > 0) then
```
HelinWei-NOAA, line 2301: @SamuelTrahanNOAA Is it better if we let model stopped under these two cases: (1) use_lake_model=o while lkm is not (2) lkm=o while use_lake_model is not

HelinWei-NOAA: In other words use_lake_model, lkm, and iopt_lake should not be totally independent

HelinWei-NOAA (approved):

ccpp/data/GFS_typedefs.F90
SamuelTrahanNOAA: The `use_lake_model(:)` is an array, not a single flag. It replaces `use_flake`. The value of `use_lake_model(i)` is `lkm` where there are lakes, and `0` elsewhere.

HelinWei-NOAA: Got it. If there is no lake fraction/depth data available, and lkm is set to 1 or 2, will the model be stopped with the error message?

SamuelTrahanNOAA: No. That would require extra logic at the workflow level. The workflow would have to change LKM based on whether lake data is available. It is better for the model to be able to run without lake data.

HelinWei-NOAA: Can we add a logic in the model level when reading the oro file? Otherwise it will be misleading. Sometimes the model may break up running the lake option without the lake data.

SamuelTrahanNOAA: Running LKM=1 or 2 without lake data isn't a problem anymore. What broke it was race conditions in threads. That has been fixed.

I'm not sure what you're asking me to do. I don't want the model to abort if lake data is missing because it would break workflows.

HelinWei-NOAA: Okay. I understand now. Thanks.

grantfirl:
@SamuelTrahanNOAA https://github.com/ufs-community/ccpp-physics/pull/26 is merged so you can update this branch in prep for its merge.

jkbk2004 (approved):

jkbk2004:
@DusanJovic-NOAA @junwang-noaa @BrianCurtis-NOAA we can merge in this pr with final approvals.

ccpp/config/ccpp_prebuild_config.py
```diff
@@ -1,3 +1,4 @@
+
```
DusanJovic-NOAA: Is this blank line added intentionally?

SamuelTrahanNOAA: Wow! No, that's quite a bug; the kernel would have used sh to execute that script. Thank you for catching it. I have removed the blank line.

junwang-noaa:
@SamuelTrahanNOAA Now we are in the transition to move the restart file writing to write grid component. I saw you made changes in FV3GFS_io.F90, corresponding changes need to go to FV3GFS_restart_io.F90 so that the write grid comp restart file writing feature can be maintained. I think you can do that in a PR after this, is that OK? Thanks

DusanJovic-NOAA (approved):

SamuelTrahanNOAA:
> corresponding changes need to go to FV3GFS_restart_io.F90

Yes, definitely. The smoke and dust restart logic isn't in there either. We'd need alternative tests for the hrrr_control and rrfs_smoke_conus13km_hrrr_warm that use the grid component to write the restart. Can you tell me how to enable that feature?

junwang-noaa:
You can set "quilting_restart: .true." in the model_configure.

junwang-noaa (approved):

PR #610, opened 2022-12-04
HAFSv1 related developments and moving-nesting code restructure

### LINKED ISSUE #182: Modularization of moving nest and storm tracker functionality.
**Is your feature request related to a problem? Please describe.**
Modularization of moving nest and storm tracker functionality

**Describe the solution you'd like**
GFDL requests a code update with the following features:

1. Take moving nested grid code and move it into its own component (at the same level as atmos_cubed_sphere). 
2. Physics code must be moved out of FV3 or its submodules.
3. Access FV3 exclusively through APIs defined in atmosphere.F90

**Describe alternatives you've considered**
AOML and GFDL agreed to initial changes for PR 179, and these additional steps to be accomplished in an upcoming release.

**Additional context**
Additional comments from GFDL:

You may consider promoting the moving nested grid to be a separate 
component, on the same level as atmos_cubed_sphere. This would be similar to the implementation of stochastic physics. In this case, the moving nest implementation would still be dependent upon FV3 but not necessarily a part of it, and communicates with FV3 through the APIs defined in atmosphere.F90. Additionally, the vortex tracker could also be separate from the moving nest code.

While many others have implemented new capabilities within #ifdef/#endif blocks, there is a tendency to overuse this approach to mask out features which may have limited adoption.  Along with the modularization recommended above, it may be acceptable to include moving nest options directly within the model without an #ifdef block.

Please note that the vast majority of FV3-based model users will not be concerned with the moving nested grids. This is especially true for our many users outside of the UFS (NASA, GFDL, NCAR, etc.) and we need to ensure the unified codebase is useful for all of our users.

Other notes:

Have computational performance (and repro?) baselines been performed to ensure other applications have not seen degraded performance?

If the simplified version of Tim’s GFDL Hurricane Tracker is useful outside of this context maybe it is better to separate it from the moving nest code

Will GFDL be responsible for maintaining the moving nested-grid code once the merge is complete? If so, it would be useful if AOML could provide some sort of documentation to explain how this works.

### LINKED ISSUE #1515: Bring in HAFSv1 related developments
## Description
Merge HAFSv1 related developments back to ufs-weather-model develop branch.

## Solution
HAFSv1 related developments has been tested with the HAFS application. Will merge related branches/submodules back to corresponding authoritative branches.

## Alternatives
N/A.

## Related to
Directly reference any issues or PRs in this or other repositories that this is related to, and describe how they are related. Example:
- addressed by ufs-community/ufs-weather-model/pull<pr_number>
- related to NOAA-EMC/fv3atm/pull/610
- related to NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/227
- related to NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/182
- related to NOAA-EMC/HYCOM-src/pull/3

BinLiu-NOAA:
## Description

Merge HAFSv1 related developments back to FV3ATM develop branch, including:
- Moving nesting code restructure (basically moving the moving nesting related source code from FV3 dycore into FV3ATM) and porting moving nesting related code to use fms2_io (from @wramstrom). More details can be seen from NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/227.
- Add the upoff as a namelist option for FV3dycore, which is recommended by GFDL colleagues and implemented by (@BijuThomas-NOAA, @wramstrom, @BinLiu-NOAA, et al.)
- Fix the storm-following moving-nesting FV3 internal tracker high-resolution tracker output time format (from @BinLiu-NOAA).
- For HAFS FV3ATM-HYCOM coupling, update to only downscale SST from FV3ATM parent (received from HYCOM through CMEPS) into its nested domain on the ocean coupling time step. (Mainly from @BinLiu-NOAA, @wramstrom, and @danrosen25).
- Add the tc_pbl option for the GFS TKE EDMF PBL scheme (mainly from @chxmoonlight and @AndrewHazelton, from AOML/HRD). More details can be seen from ufs-community/ccpp-physics/pull/27.
- Add some additional HAFS related CCPP physics suites (mainly from @ChunxiZhang-NOAA).

### Issue(s) addressed
- fixes ufs-community/ufs-weather-model/issues/1515
- fixes NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/182

## Testing
Regression testing were conducted at the ufs-weather-model level and the HAFS application level. Only HAFS related regression tests get changing results (which is expected). All other regression tests got identical results.

## Dependencies
- waiting on NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/227
- waiting on ufs-community/ccpp-physics/pull/27
- related to NOAA-EMC/HYCOM-src/pull/3
- related to ufs-community/ufs-weather-model/pull/1518


DusanJovic-NOAA:
There is no need to have CMakeLists.txt in moving_nest directory and build it as a separate project/library, all that's needed is something like this in fv3atm:
```
if(MOVING_NEST)                                                                  
  list(APPEND _fv3atm_defs_private MOVING_NEST MOIST_CAPPA USE_COND)             
  if(DEBUG)                                                                      
    list(APPEND _fv3atm_defs_private DEBUG)                                      
  endif()                                                                        
  if(GFS_TYPES)                                                                  
    list(APPEND _fv3atm_defs_private GFS_TYPES)                                  
  endif()                                                                        
  if(INTERNAL_FILE_NML)                                                          
    list(APPEND _fv3atm_defs_private INTERNAL_FILE_NML)                          
  endif()                                                                        
  if(ENABLE_QUAD_PRECISION)                                                      
    list(APPEND _fv3atm_defs_private ENABLE_QUAD_PRECISION)                      
  endif()                                                                        
  if(32BIT)                                                                      
    list(APPEND _fv3atm_defs_private OVERLOAD_R4 OVERLOAD_R8)                    
  endif()                                                                        
                                                                                 
  list(APPEND moving_nest_srcs                                                        
    moving_nest/bounding_box.F90                                                 
    moving_nest/fv_tracker.F90                                                   
    moving_nest/fv_moving_nest.F90                                               
    moving_nest/fv_moving_nest_main.F90                                          
    moving_nest/fv_moving_nest_physics.F90                                       
    moving_nest/fv_moving_nest_types.F90                                         
    moving_nest/fv_moving_nest_utils.F90                                         
  )                                                                              
endif() 
```
and then add `moving_nest_srcs` to `add_library(fv3atm ....)`. This conditionally adds 7 'moving nest' files to fv3atm library. There is no need to maintain separate compiler flags in moving_nest/cmake, no need to keep copies of FindESMF.cmake and FindNetCDF.cmake (esmf and netcdf are already found), no need to build and install separate library just for these 7 files etc.

BinLiu-NOAA:
> There is no need to have CMakeLists.txt in moving_nest directory and build it as a separate project/library, all that's needed is something like this in fv3atm:

Thanks for the comments/suggestions, @DusanJovic-NOAA! I am including @wramstrom in this conversation, since he is the one who implemented this. Thanks!

wramstrom:
@DusanJovic-NOAA I agree that this is a clunky cmake setup for the moving_nest directory.  I began with a cmake setup of just adding the new files to FV3/CMakeLists.txt, however the model run did not produce bitwise identical results.  After some investigation, I discovered that the atmos_cubed_sphere configuration uses compiler option "-fp_model source" while the FV3 directory uses "-fp-model consistent".   

So, I copied the cmake setup from atmos_cubed_sphere for the moving_nest directory to enable the same settings.  

1.  Does anyone have comments on the settings for "-fp_model"?
2. If we want to preserve the difference in that setting, is there a simpler configuration we can use to accomplish it?

DusanJovic-NOAA:
Are results not bitwise identical between this branch and the current baseline, or they are not bitwise identical between two subsequent runs (i.e. results are not reproducible)?

wramstrom:
The results were different when compared with a baseline run.

DusanJovic-NOAA:
Ok. We should accept that because of the fp-model option change from 'source' to 'consistent' for these 7 files after files were moved from the dycore to the fv3atm, and that explains (hopefully) slightly different result. But it is important that results are reproducible from one run to the next.

wramstrom:
@DusanJovic-NOAA I removed the CMake files in the moving_nest directory and added them to the main CMakeLists.txt in FV3.  This makes the structure simpler.  Model results with the new setting "fp-model consistent" are slightly different from the old setting of "fp-model source".

DusanJovic-NOAA:
> @DusanJovic-NOAA I removed the CMake files in the moving_nest directory and added them to the main CMakeLists.txt in FV3. This makes the structure simpler. Model results with the new setting "fp-model consistent" are slightly different from the old setting of "fp-model source".

Thanks @wramstrom . This is now much cleaner, and easier to maintain. Now that `moving_nest` directory is no longer (unconditionally) added with `add_subdirectory`, but the 7 files are conditionally added to the fv3atm target directly, I think there's no need to have `#ifdef MOVING_NEST` in each of the moving nest modules.

CMakeLists.txt
```diff
@@ -39,10 +40,46 @@ if(NOT PARALLEL_NETCDF)
   list(APPEND _fv3atm_defs_private NO_PARALLEL_NETCDF)
 endif()
 
+
```
junwang-noaa: remove these two empty lines?

BinLiu-NOAA: Done. Thanks, @junwang-noaa!

moving_nest/fv_moving_nest.F90
```diff
@@ -0,0 +1,2517 @@
+!***********************************************************************
+!*                   GNU Lesser General Public License
+!*
+!* This file is part of the FV3 dynamical core.
```
junwang-noaa: Not sure if you want to update this section in the future, maybe make it specific to moving nest dycore?

wramstrom: Good catch -- I have updated this with the wording that is used in atmos_model.F90: "This file is a part of fvGFS."

junwang-noaa:
@BinLiu-NOAA @wramstrom Thanks for making the code changes. The code changes aer clean. One more question, since the moving nest calls are now moved to atmo_model.F90, will this have big impact to the results since previously the moving nest subroutines are called within dycore?

wramstrom (approved):

wramstrom:
@junwang-noaa The relocation of the moving_nest directory did not change the results when compiled with the exact same compiler options; forecast results were bitwise identical.  However, the settings for "fp-model" are different in the FV3 directory vs. the atmos_cubed_sphere model.  We have moved over to use the FV3 settings for the moving_nest directory, so there are small differences in the results due to different rounding of floating point numbers.

junwang-noaa (approved):

grantfirl:
@jkbk2004 @BinLiu-NOAA https://github.com/ufs-community/ccpp-physics/pull/27 has been merged. It looks like https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/227 needs to be merged before .gitmodules is reverted and submodule pointers are updated here.

jkbk2004:
@grantfirl @BinLiu-NOAA yes, I asked https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/227 to be merged.

jkbk2004:
@BinLiu-NOAA the cubed sphere pr was merged. can you update the submodule pointer?

DusanJovic-NOAA (approved):

BinLiu-NOAA:
@jkbk2004, submodule updated now. Thanks!

PR #628, opened 2023-02-27
Lightning threat indexes

SamuelTrahanNOAA:
## Description

Calculates the lightning threat index product from HRRR, inside CCPP and sends it to UPP for output.

The algorithm was originally ported from HRRR to the GFDL_atmos_cubed_sphere NGGPS diagnostics, using the w variable in both implementations. I've moved it to CCPP's maximum hourly diagnostics, but it still needs w in the formulation, so I've added the dynamical core's w variable as wgrs to match the u & v winds which are ugrs and vgrs.

### Issue(s) addressed

https://github.com/ufs-community/ufs-weather-model/issues/1632
https://github.com/NOAA-EMC/UPP/issues/672

## Testing

A 3km test case from a time that had some lightning was run and compared to the HRRR output. Developers of the scheme found the results to be comparable. There'll be more extensive testing after this is merged and running in parallels.

Hera intel.

No, because there are no 3km test cases in ufs-weather-model for the RRFS, and the 13km are unable to produce lightning. If someone has an idea for how to fix this, please comment here.

Yes, hera.intel, but I'll run again soon after updating to the top of develop.

Many regression test results may change due to updating to the latest UPP and adding fields to regional output.

As you wish.

## Dependencies

- waiting on https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/243
- waiting on https://github.com/ufs-community/ccpp-physics/pull/42

- ~~waiting on https://github.com/NOAA-EMC/UPP/pull/643~~
- ~~waiting on https://github.com/NOAA-EMC/UPP/pull/646~~
- waiting on https://github.com/NOAA-EMC/UPP/pull/673


ccpp/data/GFS_typedefs.F90
```diff
@@ -2027,6 +2036,7 @@ subroutine statein_create (Statein, IM, Model)
     allocate (Statein%pgr    (IM))
     allocate (Statein%ugrs   (IM,Model%levs))
     allocate (Statein%vgrs   (IM,Model%levs))
+    allocate (Statein%wgrs   (IM,Model%levs))
```
junwang-noaa: Is this wgrs only used when the Model%lightning_threat is .true.? If yes, may we only allocate it when  Model%lightning_threat is .true.?

SamuelTrahanNOAA: Yes. The atmos_cubed_sphere will have to know about that, though. It is updated by the dynamical core.

ccpp/data/GFS_typedefs.F90
```diff
@@ -4977,6 +5000,13 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
     Model%restart          = restart
     Model%lsm_cold_start   = .not. restart
     Model%hydrostatic      = hydrostatic
+
+    if(Model%hydrostatic .and. Model%lightning_threat) then
```
junwang-noaa, line 5116: I assume with this setting, we do not need to pass hydrostatic to ccpp phys diagnostic subroutine?

SamuelTrahanNOAA: I'm passing it anyway, as a precaution. In the hydrostatic case, the three lightning threat indexes are filled with 0, and wgrd is never accessed.

junwang-noaa: My understanding is that the three lightning threat indexes will not be allocated, please see line 7156, and the if statements of filling the index arrays to zero in ccpp physics will not be called. In other words, the lightning threat indexes will not be output to history files.

SamuelTrahanNOAA: My worry is that `hydrostatic` will change value, since this diagnostic print says it can:

```f90
      print *, 'variables modified at each time step'
      ...
      print *, ' hydrostatic       : ', Model%hydrostatic
```

junwang-noaa: I see, Thanks. It's interesting that this "hydrostatic" variable could change during integration. @ChunxiZhang-NOAA @grantfirl Can you confirm that the hydrostatic variable could change during model integration?  I do not see the code that changes hydrostatic, maybe I missed something.

ccpp/data/GFS_typedefs.F90
ChunxiZhang-NOAA: My question is: Why not use existing state variable vvl (omega pa/s) to calcuate wgrs by vvl = dp/dz*wgrs?

SamuelTrahanNOAA: I would like the developer of the CCPP code, @welewis69, to answer that question.

ccpp/data/GFS_typedefs.F90
ChunxiZhang-NOAA: I don't think "hydrostatic" variable can be changed during integration.

grantfirl: I don't think it makes sense or is possible to change between hydrostatic and non-hydrostatic during integration. I also see no evidence of hydrostatic being set anywhere outside of initialization.

ccpp/data/GFS_typedefs.F90
```diff
@@ -160,6 +160,7 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: pgr  (:)     => null()  !< surface pressure (Pa) real
     real (kind=kind_phys), pointer :: ugrs (:,:)   => null()  !< u component of layer wind
     real (kind=kind_phys), pointer :: vgrs (:,:)   => null()  !< v component of layer wind
+    real (kind=kind_phys), pointer :: wgrs (:,:)   => null()  !< w component of layer wind
```
grantfirl, line 165: Even though it's a bit more computation, I would strongly suggest just passing in omega to maximum_hourly_diagnostics rather than trying to use `w` from the dycore. It's much simpler (you don't have to worry about the logic around whether wgrs should exist and adding a dependency of this diagnostic on whether the dycore is run hydrostatically or not). There is precedence: w is calculated from omega in the following schemes: GFDL MP, MG MP, NSSL MP, Thompson MP, MYNNEDMF PBL, SHOC. If you wanted to get fancy, you could pass the calculated w out of the microphysics schemes (or otherwise check whether w has already been calculated) within maximum_hourly_diagnostics to avoid recalculating if you're really worried about avoiding duplicate calculation.

ccpp/data/GFS_typedefs.F90
welewis69: Chunxi and others, our only concern regarding using wgrs as solved for from vvl = dp/dz*wgrs would be that the lightning threat indices are based on empirical relationships derived using the non-hydrostatic vertical velocity. In FV3, is the fully non-hydrostatic w recoverable from vvl in this manner?

ccpp/data/GFS_typedefs.F90
SamuelTrahanNOAA: Re-posting @welewis69's response to a similar comment:

> Chunxi and others, our only concern regarding using wgrs as solved for from vvl = dp/dz*wgrs would be that the lightning threat indices are based on empirical relationships derived using the non-hydrostatic vertical velocity. In FV3, is the fully non-hydrostatic w recoverable from vvl in this manner?

ccpp/data/GFS_typedefs.F90
ChunxiZhang-NOAA: In FV3, non-hydyostatic w is retrievable from vvl. The only issue is vvl is smoothed with the second-order smoothing. It is smoothed once by default in dynamics.

welewis69: Thank you for the clarification, Chunxi. I appreciate it.

ccpp/data/GFS_typedefs.meta
```diff
@@ -5965,6 +5973,12 @@
   units = flag
   dimensions = ()
   type = logical
+[lightning_threat]
+  standard_name = lightning_threat_indices_enabled
```
grantfirl: We're supposed to use "do_X" for flags like this.

ccpp/data/GFS_typedefs.meta
```diff
@@ -111,6 +111,14 @@
   dimensions = (horizontal_loop_extent)
   type = real
   kind = kind_phys
+[wgrs]
+  standard_name = z_wind
```
grantfirl: CF standards have "upward_air_velocity", which we should use unless there is a good reason not to. Yes, I understand that we're using x_wind and y_wind already, but there is no z_wind in CF standards for some reason.

ccpp/data/GFS_typedefs.meta
```diff
@@ -8759,6 +8773,33 @@
   dimensions = (horizontal_loop_extent)
   type = real
   kind = kind_phys
+[ltg1_max]
+  standard_name = gsl_lightning_threat_index_1
```
grantfirl: I would remove "gsl" from this unless/until there are multiple lightning threat indices to differentiate. Same with ...2,3.

SamuelTrahanNOAA: I was matching the names in UPP, but I'm happy to remove them here.

ccpp/data/GFS_typedefs.meta
SamuelTrahanNOAA: Is there a CF name for non-hydrostatic vertical velocity?

ccpp/data/GFS_typedefs.F90
grantfirl: @SamuelTrahanNOAA @ChunxiZhang-NOAA @dustinswales So, where did we land on this issue? Are we going to keep it as-is, or simplify and be consistent with other physics?

SamuelTrahanNOAA: @welewis69 - Can you respond again?  Should we stay with the unsmoothed non-hydrostatic vertical velocity, or derive it from the smoothed omega?

ccpp/data/GFS_typedefs.F90
SamuelTrahanNOAA: For my own sanity, I'm moving "hydrostatic" up to the "basic control parameters" in the control_print.

ccpp/data/GFS_typedefs.meta
grantfirl: No, if you do a search in their [standard name table](https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html), `hydrostatic` comes back with no matches. Although, you make a good point, if it is important that the vertical velocity used in the algorithm is non-hydrostatic, it might be worth adding that to the standard name: `nonhydrostatic_upward_air_velocity`. That would remind us in the future why `w` is being done differently in this instance (and to not recalculate from omega, if that's what is necessary and we decide to stick with the current implementation).

ccpp/data/GFS_typedefs.meta
SamuelTrahanNOAA: The "gsl" parts of the names are gone.

ccpp/data/GFS_typedefs.meta
SamuelTrahanNOAA: It is unsmoothed_nonhydrostatic_upward_air_velocity

```
[wgrs]
  standard_name = unsmoothed_nonhydrostatic_upward_air_velocity
  long_name = unsmoothed non-hydrostatic upward air velocity
  units = m s-1
  dimensions = (horizontal_loop_extent,vertical_layer_dimension)
  type = real
  kind = kind_phys
```

ccpp/data/GFS_typedefs.meta
SamuelTrahanNOAA: The variable name is already quite long. I've updated the standard name:

```
[lightning_threat]
  standard_name = do_lightning_threat_index_calculations
  long_name = enables the lightning threat index calculations
  units = flag
  dimensions = ()
  type = logical
```

ccpp/data/GFS_typedefs.F90
welewis69: Thanks much for the discussion and clarification of the various issues, everyone. Two of the three indices depend on graupel fluxes, and since these fluxes tend to be concentrated in the updrafts, we really do need to use the w field without smoothing if at all possible (otherwise, we will produce fields that are simultaneously too broad and too weak).

grantfirl (approved):
@SamuelTrahanNOAA @welewis69 Thanks very much for the clarification about w. It very much helps in code reviews to understand the reasoning behind code changes, especially when the method differs from other parts of the code. This looks good to me, and thanks for the contribution!

BrianCurtis-NOAA:
Waiting on reviews here before it makes it into the commit queue for UFSWM.

ccpp/data/GFS_typedefs.F90
junwang-noaa: @SamuelTrahanNOAA Are you planning to make any changes?

io/post_fv3.F90
```diff
@@ -741,6 +744,9 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
           ncfrst(i,j) = 1.0
           bgroff(i,j) = spval
           rswinc(i,j) = spval
+          ltg1_max(i,j) = spval
+          ltg2_max(i,j) = spval
+          ltg3_max(i,j) = spval
```
WenMeng-NOAA: @SamuelTrahanNOAA I would think initializing ltg1_max, ltg2_max and ltg3_max might not be needed since they are already initialized in UPP code ALLOCATE_ALL.f , see [here](https://github.com/NOAA-EMC/UPP/blob/develop/sorc/ncep_post.fd/ALLOCATE_ALL.f#L424)

SamuelTrahanNOAA: Why are the other fields initialized here?

WenMeng-NOAA: @SamuelTrahanNOAA Those were set before we added all interval variable initializing in ALLOCATE.f.

io/post_fv3.F90~
WenMeng-NOAA, line 1: @SamuelTrahanNOAA Somehow this file was inadvertently committed again.

SamuelTrahanNOAA: There is a missing `*` in `.gitignore`. It should be `*~` not `~`

ccpp/data/GFS_typedefs.F90
SamuelTrahanNOAA: The hydrostatic check is removed from the ccpp/physics code, and the hydrostatic print is in the right section.

io/post_fv3.F90
SamuelTrahanNOAA: Those = spval are gone now

io/post_fv3.F90~
SamuelTrahanNOAA: I added the missing `*` and git is ignoring emacs autosave files now.

WenMeng-NOAA (approved):
The changes for the in-line post part look good to me.

grantfirl:
@jkbk2004 @SamuelTrahanNOAA https://github.com/ufs-community/ccpp-physics/pull/42 is merged. Looks like https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/243 needs to be merged next.

jkbk2004:
@SamuelTrahanNOAA cubed-sphere pr was merged.

jkbk2004 (approved):

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "atmos_cubed_sphere"]
   path = atmos_cubed_sphere
-  url = https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere
-  branch = dev/emc
+  url = https://github.com/SamuelTrahanNOAA/GFDL_atmos_cubed_sphere
+  branch = lightning
```
BrianCurtis-NOAA: This still need to be updated/reverted if the atmos_cubed PR was merged.

BrianCurtis-NOAA (changes requested):

SamuelTrahanNOAA:
The .gitmodules, ccpp/physics, and atmos_cubed_sphere now point to the authoritative branches.

jkbk2004:
@SamuelTrahanNOAA correct cubed-sphere hash is https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/commit/a4450fbed77aca481c50db1623454fe7ff58d8bd can you revert the change in .gitmodules as well?

SamuelTrahanNOAA:
Sorry, that was the wrong PR. I just updated this one. It should be right, now.

.gitmodules
SamuelTrahanNOAA: I have reverted it.

BrianCurtis-NOAA (approved):

PR #641, opened 2023-03-30
Update in-line post for RRFS and GFS

WenMeng-NOAA:
## Description

- Update the upp revision with the latest commit at UPP repository
- Update the in-line post read interface for RRFS, GEFS-Aerosols

### Issue(s) addressed

- fixes #<issue_number>
- fixes ufs-weather-model/issues/[1688](https://github.com/ufs-community/ufs-weather-model/issues/1688)

## Testing

## Dependencies

- waiting on noaa-emc/nems/pull/<pr_number>
- waiting on noaa-emc/fv3atm/pull/<pr_number>

### DIFF
--- a/io/post_fv3.F90
+++ b/io/post_fv3.F90
@@ -40,7 +40,7 @@ module post_fv3
 !-----------------------------------------------------------------------
 !
       use ctlblk_mod, only : komax,ifhr,ifmin,modelname,datapd,fld_info, &
-                             npset,grib,gocart_on,jsta,  &
+                             npset,grib,jsta,  &
                              jend,ista,iend, im, nsoil, filenameflat,numx
       use gridspec_mod, only : maptype, gridtype,latstart,latlast,       &
                                lonstart,lonlast
@@ -501,7 +501,8 @@ module post_fv3
 !-----------------------------------------------------------------------
 !
       use esmf
-      use vrbls4d,     only: dust, smoke, fv3dust, coarsepm
+      use vrbls4d,     only: dust, smoke, fv3dust, coarsepm, SALT, SUSO, SOOT, &
+                             WASO,no3,nh4, PP25, PP10
       use vrbls3d,     only: t, q, uh, vh, wh, alpint, dpres, zint, zmid, o3,  &
                              qqr, qqs, cwm, qqi, qqw, qqg, omga, cfr, pmid,    &
                              q2, rlwtt, rswtt, tcucn, tcucns, train, el_pbl,   &
@@ -509,7 +510,7 @@ module post_fv3
                              qqnifa, effri, effrl, effrs, aextc55, taod5503d,  &
                              duem, dusd, dudp, duwt, dusv, ssem, sssd, ssdp,   &
                              sswt, sssv, bcem, bcsd, bcdp, bcwt, bcsv, ocem,   &
-                             ocsd, ocdp, ocwt, ocsv
+                             ocsd, ocdp, ocwt, ocsv, rhomid
       use vrbls2d,     only: f, pd, sigt4, fis, pblh, ustar, z0, ths, qs, twbs,&
                              qwbs, avgcprate, cprate, avgprec, prec, lspa, sno,&
                              cldefi, th10, q10, tshltr, pshltr, albase,        &
@@ -544,19 +545,22 @@ module post_fv3
                              acgraup, graup_bucket, acfrain, frzrn_bucket,     &
                              ltg1_max, ltg2_max, ltg3_max, aodtot, ebb, hwp,   &
                              aod550,du_aod550,ss_aod550,su_aod550,oc_aod550,   &
-                             bc_aod550,maod
+                             bc_aod550,maod,                                   &
+                             dustpm10, dustcb, bccb, occb, sulfcb, sscb,       &
+                             dustallcb, ssallcb, dustpm, sspm, pp25cb, pp10cb, &
+                             no3cb, nh4cb, dusmass, ducmass, dusmass25,ducmass25
       use soil,        only: sldpth, sh2o, smc, stc, sllevel
       use masks,       only: lmv, lmh, htm, vtm, gdlat, gdlon, dx, dy, hbm2, sm, sice
       use ctlblk_mod,  only: im, jm, lm, lp1, jsta, jend, jsta_2l, jend_2u, jsta_m,jend_m, &
-                             ista, iend, ista_2l, iend_2u, ista_m,iend_m, &
+                             ista, iend, ista_2l, iend_2u, ista_m,iend_m,qmin, &
                              lsm, pt, imp_physics, spval, mpi_comm_comp, gdsdegr,  &
                              tprec, tclod, trdlw, trdsw, tsrfc, tmaxmin, theat, &
                              ardlw, ardsw, asrfc, avrain, avcnvc, iSF_SURFACE_PHYSICS,&
                              td3d, idat, sdat, ifhr, ifmin, dt, nphs, dtq2, pt_tbl, &
                              alsl, spl, ihrst, modelname, nsoil, rdaod, gocart_on,  &
                              gccpp_on, nasa_on, d2d_chem, nbin_ss, nbin_bc, nbin_oc,&
-                             nbin_du
-      use params_mod,  only: erad, dtr, capa, p1000, small
+                             nbin_du,nbin_su, nbin_no3, nbin_nh4
+      use params_mod,  only: erad, dtr, capa, p1000, small,h1, d608, pi, rd
       use gridspec_mod,only: latstart, latlast, lonstart, lonlast, cenlon, cenlat, &
                              dxval, dyval, truelat2, truelat1, psmapf, cenlat,     &
                              lonstartv, lonlastv, cenlonv, latstartv, latlastv,    &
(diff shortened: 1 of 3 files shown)


io/post_fv3.F90
```diff
@@ -679,6 +695,12 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
       dtq2 = wrt_int_state%dtp
       nphs = 2.
       dt   = dtq2/nphs
+
+      allocate(extsmoke(ista:iend,jsta:jend,lm))
```
junwang-noaa: Do we need to allocate those array for all the applications?

WenMeng-NOAA: Right now they are for RRFS only.

io/post_fv3.F90
```diff
@@ -2138,165 +2335,644 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
-                enddo
+            endif !end rdaod
+
+            if ((gocart_on .or. gccpp_on) .and. d2d_chem) then
+              do K = 1, nbin_du
+                if ( K == 1) VarName='duem001'
+                if ( K == 2) VarName='duem002'
+                if ( K == 3) VarName='duem003'
+                if ( K == 4) VarName='duem004'
+                if ( K == 5) VarName='duem005'
```
DusanJovic-NOAA: What will be name for K>5? 
Instead of all these if tests it is simpler to:
`write(VarName, '(A,I3.3)')  'duem', K
`
or something like that

io/post_fv3.F90
```diff
@@ -86,7 +86,8 @@ subroutine post_run_fv3(wrt_int_state,grid_id,mype,mpicomp,lead_write, &
 
       grib      = "grib2"
       gridtype  = "A"
-      nsoil     = 4
+      !nsoil     = 4
```
DusanJovic-NOAA: Remove this line if it's not used

io/post_fv3.F90
```diff
@@ -799,9 +821,9 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
 !$omp& shared(smstav,sfcevp,acsnow,acsnom,qz0,uz0,vz0,maxrhshltr,minrhshltr)
       do j=jsta_2l,jend_2u
         do i=ista_2l,iend_2u
-          smstav(i,j) = spval
+          !smstav(i,j) = spval
           sfcevp(i,j) = spval
-          acsnow(i,j) = spval
+          !acsnow(i,j) = spval
```
DusanJovic-NOAA: Remove these two lines if they are note needed anymore. And two variables from omp directive

io/post_fv3.F90
WenMeng-NOAA: nbin_du is set as 5 in UPP CTLBLK.f
`integer, parameter :: nbin_du = 5`

WenMeng-NOAA:
@DusanJovic-NOAA I updated io/post_fv3.F90 based on your comment.

WenMeng-NOAA:
@junwang-noaa and @DusanJovic-NOAA Any actions are needed from my side? Thanks!

io/post_fv3.F90
junwang-noaa: This may increase the memory requirement, I want to confirm for other applications these arrays will not be allocated? I do not see an if statement to limit the allocation just for RRFS.

WenMeng-NOAA: @junwang-noaa I added changes based on your comment. Please let me know any tweaking is needed.

io/post_fv3.F90
```diff
@@ -2138,164 +2335,641 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
               enddo
             endif
 
-            ! inst  cloud top pressure
-            if(trim(fieldname)=='prescnvclt') then
```
junwang-noaa, line 2142: @WenMeng-NOAA So we don't output inst/time averaged cloud top/mid/bottom pressure and temperature any more?

io/post_fv3.F90
WenMeng-NOAA: Done!

io/post_fv3.F90
WenMeng-NOAA: Done!

io/post_fv3.F90
WenMeng-NOAA: Done!

io/post_fv3.F90
WenMeng-NOAA: @junwang-noaa These variables are still kept in the code. There are a lot of new variables inserted before them.

junwang-noaa (approved):

io/post_fv3.F90
```diff
@@ -2138,164 +2335,641 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
+                    do i=ista, iend
+                      maod(i,j) = arrayr42d(i,j)
+                      if( abs(arrayr42d(i,j)-fillValue) < small) maod(i,j) = spval
+                    enddo
+                  enddo
+                endif
+
+            endif !end gocart_on
+
+            ! inst  cloud top pressure
```
WenMeng-NOAA, line 2799: @junwang-noaa Please see reading of cloud related variables starting here.

io/post_fv3.F90
```diff
@@ -2138,164 +2335,641 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
+                  buf2(i,j) = arrayr42d(i,j)
+                  if(abs(arrayr42d(i,j)-fillvalue)<small) buf2(i,j) = spval
                 enddo
               enddo
             endif
+            ! !$omp parallel do default(none) private(i,j) shared(jsta,jend,ista,iend,spval,buf,buf2,acsnow)
+             do j=jsta,jend
+               do i=ista, iend
+                 if(buf(i,j)<spval .and. buf2(i,j)<spval) then
+                   acsnow(i,j) = buf(i,j) + buf2(i,j)
```
DusanJovic-NOAA: What are the values in buf and buf2 arrays? I guess they should be `accswe_land` and `accswe_ice`. If I understand correctly this i,j loop will be executed for every iteration of `do n=1, ncount_field` loop since it's not inside an if test that checks the value of `fieldname`. Is that correct?

io/post_fv3.F90
```diff
@@ -2138,164 +2335,641 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
             endif
-            ! time averaged middle cloud bottom pressure
-            if(trim(fieldname)=='pres_avemcb') then
-              !$omp parallel do default(none) private(i,j) shared(jsta,jend,ista,iend,spval,pbotm,arrayr42d,fillValue)
-              do j=jsta,jend
-                do i=ista, iend
-                  pbotm(i,j) = arrayr42d(i,j)
-                  if( abs(arrayr42d(i,j)-fillValue) < small) pbotm(i,j) = spval
-                enddo
+            !!$omp parallel do default(none) private(i,j) shared(jsta,jend,ista,iend,spval,buf,buf2,sndepac)
```
DusanJovic-NOAA: Is `!!` intended here?

io/post_fv3.F90
WenMeng-NOAA: @DusanJovic-NOAA That's right. Do you have any suggestions on this?
Adding if(trim(fieldname)=='accswe_ice' .and. trim(fieldname)=='accswe_ice') before i,j loop for calculating acsnow?

io/post_fv3.F90
WenMeng-NOAA: @DusanJovic-NOAA Thanks for catching it. Will delete one !.

io/post_fv3.F90
WenMeng-NOAA: @DusanJovic-NOAA Do you think calculation of acsnow and andepac should be removed the line outside of ncount_field loop?

DusanJovic-NOAA: That if test will not work because fieldname can not be both `accswe_land` and `accswe_ice` at the same time.

Also this logic depends on the order of the fields in the bundle, which I do not think is very robust. What if `accswe_land` and `accswe_ice` are not next to each other and for whatever reason other fields that also assign the values to buf or buf2, like `snacc_ice` are in between these two.

DusanJovic-NOAA: Ideally this computation should be done somewhere in the model, not here. This routine should only copy data from 'model arrays' to 'post arrays', nothing else.

WenMeng-NOAA: > That if test will not work because fieldname can not be both `accswe_ice` and `accswe_ice` at the same time.
> 
> Also this logic depends on the order of the fields in the bundle, which I do not think is very robust. What if `accswe_ice` and `accswe_ice` are not next to each other and for whatever reason other fields that also assign the values to buf or buf2, like `snacc_ice` are in between these two.

@DusanJovic-NOAA You are right. How about moving calculation of acsnow and andepac outside of ncount_field loop (e.g. after line 4656) and also allocate another two bufs for snacc_land and snacc_ice?

WenMeng-NOAA: > Ideally this computation should be done somewhere in the model, not here. This routine should only copy data from 'model arrays' to 'post arrays', nothing else.

@DusanJovic-NOAA I will pass your comment to RRFS developers from GSL for future enhancement.

DusanJovic-NOAA: > > That if test will not work because fieldname can not be both `accswe_ice` and `accswe_ice` at the same time.
> > Also this logic depends on the order of the fields in the bundle, which I do not think is very robust. What if `accswe_ice` and `accswe_ice` are not next to each other and for whatever reason other fields that also assign the values to buf or buf2, like `snacc_ice` are in between these two.
> 
> @DusanJovic-NOAA You are right. How about moving calculation of acsnow and andepac outside of ncount_field loop (e.g. after line 4656) and also allocate another two bufs for snacc_land and snacc_ice?

That should work. And instead of buf and buf2 please use arrays named accswe_land, accswe_ice, snacc_land, snacc_ice.

junwang-noaa: You may want to use two logic 2D variables to mark if the two fields are set up from the model. Also I think final fields can't be set up as: ice+land because the two fields can't overlap with each other, you may end up with the full field values >spval. See line 2356 and 2536.

WenMeng-NOAA:
@DusanJovic-NOAA Please let me know your comment on the commit 78cfb35.

io/post_fv3.F90
```diff
@@ -3384,6 +4625,51 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
+                 else
+                   acsnow(i,j) = spval
+                 endif
+               enddo
+             enddo
+
+            !$omp parallel do default(none) private(i,j) shared(jsta,jend,ista,iend,spval,snacc_ice,snacc_land,sndepac)
+            do j=jsta,jend
+              do i=ista, iend
+                if(snacc_ice(i,j)<spval .and. snacc_land(i,j)<spval) then
```
junwang-noaa: This if statement may not be true for any land point or ice point.

io/post_fv3.F90
```diff
@@ -3384,6 +4625,51 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
+                endif
+              enddo
+            enddo
+
+      deallocate(extsmoke)
+      deallocate(extdust)
+      deallocate(buf)
+      deallocate(buf2)
+      deallocate(buf3)
+      deallocate(buf4)
```
DusanJovic-NOAA: There are no buf, buf2, buf3 and buf4 anymore.

io/post_fv3.F90
DusanJovic-NOAA: To prevent these kid of bugs in the future I think almost the entire `ncount_field` loop should be rewritten using 
```

if(trim(fieldname)==......) then

else if(trim(fieldname)==......) then

end if
```

or select case statement instead of separate if tests

Also there is a code at the end of  `ncount_field` loop that is executed for every field, for example loop that computes PM concentrations. I do not think that should be done in a loop that is executed hundreds of times.

io/post_fv3.F90
WenMeng-NOAA: @DusanJovic-NOAA Thanks for catching that.

io/post_fv3.F90
```diff
@@ -3085,94 +3743,677 @@ subroutine set_postvars_fv3(wrt_int_state,grid_id,mype,mpicomp)
...
+           if (nasa_on) then
+!      Surface PM10 concentration
+           dusmass(i,j)=pp10(i,j,l,1)*RHOMID(i,j,l)  !ug/m3
+!      Surface PM25 concentration       
+           dusmass25(i,j)=pp25(i,j,l,1)*RHOMID(i,j,l)  !ug/m3
+
+!      PM10 column
+           ducmass(i,j)=pp10cb(i,j)
+!      PM25 column
+           ducmass25(i,j)=pp25cb(i,j)
```
DusanJovic-NOAA: dusmass, dusmass25, ducmass and ducmass25 are computed both in if (gocart_on .or. gccpp_on) and if (nasa_on). 
If both gocart_on and nasa_on are true then second compuation will overwrite the first one. Can they both be true at the same time? If not then this should be `if ... else if .. end if` not two separate if blocks.

io/post_fv3.F90
WenMeng-NOAA: @junwang-noaa According to model developer from GSL, that's the intention for calculating acsnow and sdepac for RRFS. Can we use this a workaround in in-line post read interface and wish GSL change inside model for directly outputting these two variables?  Should I open ufs-weather-model or fv3atm issue for this?

DusanJovic-NOAA:
Not related to this PR, but since I'm looking at the code I thought I should mention, there are many OMP directives in which `im` is listed as shared variable but it is not used in the code at all, for example
```
!$omp parallel do default(none) private(i,j) shared(jsta,jend,im,spval,pt,pd,pint,ista,iend)
      do j=jsta,jend
        do i=ista,iend
          pd(i,j)     = spval
          pint(i,j,1) = pt
        end do
      end do
```
Also, code indentation is terrible, it is very difficult to reason about the code and understand the logic with inconsistent indentation. This file (post_fv3.F90) requires some serious clean up.

io/post_fv3.F90
WenMeng-NOAA: @DusanJovic-NOAA It could be both gocart_on and nasa_on are true. In that case it is intention second calculation overwrite the first one. @zhanglikate Can you confirm that?

io/post_fv3.F90
junwang-noaa: Yes. At this time if you want to make the change, I think you need to have if/else statements:
if(snacc_ice(i,j)<spval ) then
sndepac(i,j) = snacc_ice(i,j)
else if ( snacc_land(i,j) < spval) then
sndepac(i,j) = snacc_land(i,j)
else
sndepac(i,j) = spval
endif

io/post_fv3.F90
zhanglikate: gocart_on (or gccpp_on),  nasa_on, will never be  true at the same time since we can only run  one chemical model at one time. Thanks. @WenMeng-NOAA @DusanJovic-NOAA

io/post_fv3.F90
WenMeng-NOAA: @EricJames-NOAA How do you think about @junwang-noaa 's suggestion?

io/post_fv3.F90
WenMeng-NOAA: @zhanglikate Thanks for clarifying. I will change based on @DusanJovic-NOAA 's suggestion` if ... else if .. end if`

WenMeng-NOAA:
> Not related to this PR, but since I'm looking at the code I thought I should mention, there are many OMP directives in which `im` is listed as shared variable but it is not used in the code at all, for example
> 
> ```
> !$omp parallel do default(none) private(i,j) shared(jsta,jend,im,spval,pt,pd,pint,ista,iend)
>       do j=jsta,jend
>         do i=ista,iend
>           pd(i,j)     = spval
>           pint(i,j,1) = pt
>         end do
>       end do
> ```
> 
> Also, code indentation is terrible, it is very difficult to reason about the code and understand the logic with inconsistent indentation. This file (post_fv3.F90) requires some serious clean up.

@DusanJovic-NOAA We could submit another PR to clean up these non-fatal bugs/typos later.

DusanJovic-NOAA:
> > Not related to this PR, but since I'm looking at the code I thought I should mention, there are many OMP directives in which `im` is listed as shared variable but it is not used in the code at all, for example
> > ```
> > !$omp parallel do default(none) private(i,j) shared(jsta,jend,im,spval,pt,pd,pint,ista,iend)
> >       do j=jsta,jend
> >         do i=ista,iend
> >           pd(i,j)     = spval
> >           pint(i,j,1) = pt
> >         end do
> >       end do
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Also, code indentation is terrible, it is very difficult to reason about the code and understand the logic with inconsistent indentation. This file (post_fv3.F90) requires some serious clean up.
> 
> @DusanJovic-NOAA We could submit another PR to clean up these non-fatal bugs/typos later.

Sure. Like I said, not related to this PR.

io/post_fv3.F90
EricJames-NOAA: @WenMeng-NOAA Yes I think that's a good suggestion!  I think that code change makes sense.

WenMeng-NOAA: @EricJames-NOAA Thanks for clarifying. I will add changes based on @junwang-noaa 's comment.

WenMeng-NOAA:
@DusanJovic-NOAA @junwang-noaa @zhanglikate @EricJames-NOAA Please let me know your comments on my latest commit 246acf2.

jkbk2004:
tests are all done at weather model pr: https://github.com/ufs-community/ufs-weather-model/pull/1690 @DusanJovic-NOAA  @zhanglikate @EricJames-NOAA Please, go ahead for final reviews and approvals. @WenMeng-NOAA do you think we need to create an issue to follow up Dusan's comments to clean up codes?

DusanJovic-NOAA (approved):

WenMeng-NOAA:
> tests are all done at weather model pr: [ufs-community/ufs-weather-model#1690](https://github.com/ufs-community/ufs-weather-model/pull/1690) @DusanJovic-NOAA @zhanglikate @EricJames-NOAA Please, go ahead for final reviews and approvals. @WenMeng-NOAA do you think we need to create an issue to follow up Dusan's comments to clean up codes?

@jkbk2004 Yes, we can create an issue for that at fv3atm repos. Do you want me to create or you will do it? Thanks!

jkbk2004:
@WenMeng-NOAA we can resolve the conversations and merge in this pr. can you resolve?

jkbk2004:
> > tests are all done at weather model pr: [ufs-community/ufs-weather-model#1690](https://github.com/ufs-community/ufs-weather-model/pull/1690) @DusanJovic-NOAA @zhanglikate @EricJames-NOAA Please, go ahead for final reviews and approvals. @WenMeng-NOAA do you think we need to create an issue to follow up Dusan's comments to clean up codes?
> 
> @jkbk2004 Yes, we can create an issue for that at fv3atm repos. Do you want me to create or you will do it? Thanks!

WenMeng-NOAA:
> @WenMeng-NOAA we can resolve the conversations and merge in this pr. can you resolve?

Yes, all conversations are solved.

jkbk2004:
@WenMeng-NOAA I will let you create an issue to clean up code. merging the pr now.

PR #69, opened 2020-02-19
Add no_nsst CCPP suites

climbfuji:
Based on #67, but use different names (no_nsst instead of no_NST).

Associated PR: https://github.com/ufs-community/ufs-weather-model/pull/61 - see there for testing.

PR #654, opened 2023-05-03
Few small fixes to Thompson MP

grantfirl:
## Description

This PR includes 3 small fixes to Thompson microphysics:

The minimum size of snow was increased (some months ago) but it was noted that the min size of snow is used to determine the upper-most size bin of cloud ice. It was set to 5 times the min size of snow, which is exceedingly large for the cloud ice category. Therefore, the constant was reduced to 2 - making cloud ice largest size bin of 600 microns which is plenty big enough, giving more resolution to all the ice size bins.

In testing RRFS with @ericaligo-NOAA and @RuiyuSun for a specific case study, it was noted that rain was evaporating with a large rate producing a temperature tendency of 27C/hour. We traced the issue to an extremely large number of rain drops that appears to be produced by colliding rain and graupel - which has an explicit drop breakup (splash if you will) from the collisions making 5 drops per collision. This was forcing the rain (formed by melting graupel) to tends towards 500 microns (0.5 mm) which is too small. So the constant used for collisions was reduced to 1.5 in an effort to keep rain larger (thus reducing evaporation).

There was an oversight in the code for melting snow and graupel. When RH is below saturation, the wetbulb temperature could be below 0C in which case ice does not melt. Instead, the ice (snow/graupel) will sublimate first until wetbulb temperature goes above melting. So the code fix now enforces that either melting occurs or sublimation, not both.

Note: This PR also updates the ccpp/framework submodule pointer. This has no effect on any results.

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/1734 for testing

## Dependencies

https://github.com/ufs-community/ccpp-physics/pull/69


DusanJovic-NOAA (approved):

BrianCurtis-NOAA (approved):

gthompsnJCSDA:
It might not matter, but just an alert that the header text of this PR was not updated to align with PR 69.  The first of the four bulleted items above was removed as it was found to be correct in original code so no changes were needed for the rain evaporation item above.

grantfirl:
> 

Fixed.

zach1221:
@jkbk2004 , Grant has updated the pointer and reverted the gitmodule url, following the ccpp-physics PR [#69](https://github.com/ufs-community/ccpp-physics/pull/69) merge. Can you please go ahead to merge this PR?

PR #660, opened 2023-05-25
Refactor fv3atm history & restart to reduce redundant code. Add rrfs-sd and clm lake to quilt restart.

### LINKED ISSUE #658: Quilt Restart doesn't support the CLM Lake Model or RRFS-SD Scheme
## Description
The quilt restart cannot handle CLM lake data or RRFS-SD. Sadly, I must admit, this bug happened because I didn't know there was a quilt restart capability when I was adding those schemes.

### To Reproduce:
Enable quilt restart with a configuration that uses the CLM lake model or RRFS.

### LINKED ISSUE #659: Having three versions of Sfcprop grid copy routines for sfc restart is error-prone.
## Description
There are three giant loops to copy data to or from Sfcprop. These loops are almost, but not quite, identical. This can lead to read vs. write and quilt vs. non-quilt restart behaving differently. For example, a careless developer may forget to implement the CLM Lake model and RRFS-SD in the quilt restart (see #658).

## Solution
Someone should refactor the surface restart subroutines to eliminate redundant code.

## Alternatives
Ignore the problems.

SamuelTrahanNOAA:
## Description

This is a major refactoring of the restart and history I/O code, which also adds CLM Lake and RRFS-SD to the quilting restart. It eliminates code duplication between read vs. write and quilt vs. non-quilt. Most notably, it merges the giant `Sfcprop` copy loops for sfc read, non-quilt write, and quilt write into a single function. With these changes, the code will be easier to maintain, and read vs. write bugs will be easier to find.

1. The code in `FV3GFS_io.F90` and `FV3GFS_restart.F90` has been split into several files, divided by purpose.
    - High-level API:
        - `fv3atm_restart_io.F90` = code to read or write restart files, whether by quilt or not
        - `fv3atm_history_io.F90` = code to write history files, whether by quilt or not
    - Low-level API, only accessed by those two files:
        - `fv3atm_oro_io.F90` = code that reads oro files.
        - `fv3atm_clm_lake_io.F90` = all clm lake restart I/O
        - `fv3atm_rrfs_sd_io.F90` = all rrfs-sd restart I/O and emissions reading.
        - `fv3atm_sfc_io.F90` = all other sfc file restart I/O
        - `fv3atm_common_io.F90` = common utilities.
2. The quilt restart code can read and write CLM Lake and RRFS-SD scheme data.
3. `FV3GFS` has been changed to `fv3atm` everywhere in the fv3atm repository, except output file metadata.
4. The three massive loops that copy to and from `Sfcprop` are now a single subroutine, `Sfc_io_transfer`. This way, any differences between restart read and write will stand out prominently. (They're controlled by `reading` and `warm_start` flags.)
5. Data storage is now inside derived types, eliminating the numerous module-level arrays and indexes.
7. Quilt restart variables pass an axis index list instead of axis length (fix from @DusanJovic-NOAA)
8. An unused CLM Lake axis is gone. This is needed to get the quilt restart to match, since it can't output axis variables for axes that are never used. (fix from @DusanJovic-NOAA)
9. Comments for every subroutine, function, module, and derived type
10. Removed a large block of code guarded by `#ifdef JUNK`
11. Deleted unused variables.
12. Consistent indentation in fv3*io.F90 files (emacs f90-mode default)

### Issue(s) addressed

- Closes #658
- Closes #659

New regression tests in https://github.com/ufs-community/ufs-weather-model/pull/1769

Hera intel and gnu.

Yes, there are new regression tests.

Yes, on hera.intel and hera.gnu.

Any code that writes out CLM Lake restart data will change because I removed an unneeded axis variable. This is necessary due to a limitation of the quilt restart: it can't write axis information for axes that are never used.

Done.

## Dependencies

None.


SamuelTrahanNOAA:
I'd like to ping:
- @haiqinli whose RRFS-SD code I refactored from `FV3GFS_io.F90` into `FV3GFS_rrfs_sd_io.F90`.
- @tanyasmirnova, who is the other developer involved in CLM Lake.
- @junwang-noaa who asked me to add RRFS-SD and CLM Lake restart functionality to the quilt

io/FV3GFS_sfc_io.F90
```diff
@@ -0,0 +1,1573 @@
...
+        call register_axis(Sfc_restart, 'Time', 1)
+      end if
+    else
+      call register_axis(Sfc_restart, 'xaxis_1', 'X')
+      call register_axis(Sfc_restart, 'yaxis_1', 'Y')
+      call register_axis(Sfc_restart, 'zaxis_1', dimension_length=Model%kice)
+      if (Model%lsm == Model%lsm_noah .or. Model%lsm == Model%lsm_noahmp) then
+        call register_axis(Sfc_restart, 'zaxis_2', dimension_length=Model%lsoil)
+      else if(Model%lsm == Model%lsm_ruc .and. reading) then
+        ! Possible bug. This is only defined on read, not write.
```
tanyasmirnova: @SamuelTrahanNOAA Maybe it is a bug. Should we define it on both read/write?

tanyasmirnova (commented):
@SamuelTrahanNOAA Sam, thank you very much for your excellent work to make the I/O subroutine well-organized. It was the most confusing subroutine for me, and with your changes it would be easier to understand and add variables to I/O if needed.

io/FV3GFS_sfc_io.F90
SamuelTrahanNOAA: If it is a bug, then the other half of the bug is in `Sfc_io_register_3d_fields` where lsm_ruc registers different fields.

SamuelTrahanNOAA: I have an update on this. Tanya and I have probably narrowed down the reason for this discrepancy between read and write. It looks like there's no bug. I'm going to clean up that bit of code and make sure we're right about its bug-free nature.

SamuelTrahanNOAA:
At the behest of @DusanJovic-NOAA, I have refactored a bit more. The `fv3atm_history_io.F90` contains all history file routines, and `fv3atm_restart_io.F90` contains all restart file routines.

io/FV3GFS_sfc_io.F90
SamuelTrahanNOAA: I've added a comment explaining this oddity.

io/fv3atm_sfc_io.F90
```diff
@@ -255,7 +255,8 @@ subroutine Sfc_io_register_axes(sfc, Model, Sfc_restart, reading, warm_start)
       if (Model%lsm == Model%lsm_noah .or. Model%lsm == Model%lsm_noahmp) then
         call register_axis(Sfc_restart, 'zaxis_2', dimension_length=Model%lsoil)
       else if(Model%lsm == Model%lsm_ruc .and. reading) then
-        ! Possible bug. This is only defined on read, not write.
+        ! The RUC defines zaxis_1 for reading, so it can read a restart from a different LSM.
+        ! The lsm_ruc never writes zaxis_2.
```
tanyasmirnova: @SamuelTrahanNOAA Could you please explain in the comment what is zaxis_1 and zaxis_2? Thank you!

SamuelTrahanNOAA: I'm working my way through the code, adding comments, right now. I'll be sure to add that one.

SamuelTrahanNOAA: There is a much longer paragraph explaining the situation. How does it look now?

```f90
        call register_axis(Sfc_restart, 'zaxis_2', dimension_length=Model%lsoil_lsm)
        ! The RUC only ever writes zaxis_1, which is combined soil-ice, kice, which is 9.
        ! Other LSMs read and write zaxis_2, which is lsoil for them, and that's always 4.
        ! Defining zaxis_2 here lets RUC LSM read from a different soil vertical coordinate
        ! (lsoil_lsm). In practice, this probably won't work well since levels will be
        ! filled with zeroes. We retain this capability for historical reasons.
        ! Just make sure you only restart RUC LSM off of RUC LSM, and always have
        ! kice = lsoil = lsoil_lsm = 9 and everything will be fine.
```

tanyasmirnova: @SamuelTrahanNOAA Thank you very much for adding detailed description here. It looks good to me, although I think the sentence in bold could be confusing. 
        ! The RUC only ever writes zaxis_1, which is combined soil-ice, kice, which is 9.
        ! Other LSMs read and write zaxis_2, which is lsoil for them, and that's always 4.
        ! Defining zaxis_2 here lets RUC LSM read from a different soil vertical coordinate
        ! (lsoil_lsm). **In practice, this probably won't work well since levels will be
        ! filled with zeroes.** We retain this capability for historical reasons.
        ! Just make sure you only restart RUC LSM off of RUC LSM, and always have
        ! kice = lsoil = lsoil_lsm = 9 and everything will be fine.

I propose the following:        
        ! The RUC only ever writes zaxis_1, which is combined soil/ice vertical dimension, lsoil_lsm/kice, which is 9.
        ! Other LSMs read and write zaxis_2, which is lsoil for them, and that's always 4.
        ! Defining zaxis_2 here lets RUC LSM read from a different soil vertical coordinate
        ! (lsoil_lsm). It is needed for restart of RUC LSM from RUC LSM. This capability exists for historical reasons,
        ! because there are two sets of soil state variables: one set has lsoil=4 vertical layers (Noah LSM. NoahMP LSM), 
        ! and another set has lsoil_lsm=9 vertical levels (RUC LSM). Ideally there should be just one set of soil variables
        ! that could have different vertical dimension depending on the choice of LSM.
        ! For now: just make sure you only restart RUC LSM off of RUC LSM, and always have
        ! kice = lsoil = lsoil_lsm = 9 and everything will be fine.

SamuelTrahanNOAA: I have replaced my comment with yours and I word-wrapped it. How does it look?

tanyasmirnova: Looks great. Thank you very much, Sam.

SamuelTrahanNOAA:
It would be great if @climbfuji reviewed this, since I'm refactoring code he wrote.

tanyasmirnova (approved):

DusanJovic-NOAA (approved):

zach1221:
@tanyasmirnova would you be able to review this PR again? It appears your approval from two weeks ago dropped after a change.

SamuelTrahanNOAA:
> would you be able to review this PR again? It appears your approval from two weeks ago dropped after a change.

No, github still shows it as being approved.

Screenshot_2023-06-15_19-14-25

SamuelTrahanNOAA:
I think it wants two approvals from people with reviewer access.

zach1221:
@SamuelTrahanNOAA Yes, correct. We need one more approving review from a reviewer with write access.

BrianCurtis-NOAA (approved):

tanyasmirnova (approved):

io/fv3atm_oro_io.F90
```diff
@@ -0,0 +1,333 @@
...
+
+  end subroutine Oro_io_copy
+
+  !>@brief Destructor for Oro_io_data_type
+  subroutine Oro_io_final(oro)
+    implicit none
+    type(Oro_io_data_type) :: oro
+
+    ! This #define reduces code length by a lot
+#define IF_ASSOC_DEALLOC_NULL(var) \
```
climbfuji, line 224: Just a quick question for my understanding. This works because `name`, `var2`, ... are all pointers? Would it have been possible to create a subroutine instead to achieve the same (sorry, came back from Japan last night, my brain is not firing on all cylinders this morning).

climbfuji (approved):
This is great work @SamuelTrahanNOAA. Just one question for my understanding.

io/fv3atm_oro_io.F90
SamuelTrahanNOAA: To use a subroutine, you would need to pass a reference to a pointer, which is impossible in Fortran. Also, speaking of this construct in general, there are multiple types of pointers to be deallocated. Thus, you would need multiple subroutines. Even, hypothetically, if you could use subroutines, it would be a whole lot shorter to use a #define.

climbfuji: Thanks for the explanation. I am fine with your solution!

DusanJovic-NOAA: In Fortran arguments are passed by reference, by default.

SamuelTrahanNOAA: I don't think that is true of pointers. If you assign to a pointer, you assign to a local copy, not the one that the caller passes in. In other words, the pointer itself is the reference. You need the equivalent of this:

```C
void my_c_function(int **pointer_to_a_pointer)
{
    if(*pointer_to_a_pointer)
    {
       free(*pointer_to_a_pointer);
       *pointer_to_a_pointer = nullptr;
    }
}
```

DusanJovic-NOAA: Fortran is not C. 

In Fortran all variables are passed by reference. That includes pointers, i.e. variables with pointer attributes. So writing subroutine that accepts pointer, tests whether pointer is associated, and if it is deallocates it and nullifies it is possible.

```Fortran
program test

   real, pointer, dimension(:) :: r1d => null()

   allocate(r1d(10))
   r1d = [1,2,3,4,5,6,7,8,9,10]

   print *, 'associated ', associated(r1d)
   call dealloc_and_null_if_associated(r1d)
   print *, 'associated ', associated(r1d)

   contains

   subroutine dealloc_and_null_if_associated(ptr)
      real, pointer, dimension(:) :: ptr
      if(associated(ptr)) then
        deallocate(ptr)
        nullify(ptr)
      endif
   end subroutine dealloc_and_null_if_associated

end program test
```

SamuelTrahanNOAA: This is undergoing final testing now. I think it's too late to change it. We should wait to a later clean-up PR, and put something in fv3atm_common_io.F90.

BrianCurtis-NOAA: Final testing has not started yet. Let's make a decision, please.

dusanjovic: Of course. I am not even sure if we should make any change at all. Since, as you said, subroutine will need to be overloaded.

SamuelTrahanNOAA: In that case, we're better off with what we have now.

BrianCurtis-NOAA: OK, just making sure all is good here before we start testing. I assume it is and will resume testing.

zach1221: @SamuelTrahanNOAA if all good here, can you resolve this conversation , please?

FernandoAndrade-NOAA:
Regression tests for [1769](https://github.com/ufs-community/ufs-weather-model/pull/1769) have completed successfully. This is ready for merging. @jkbk2004

Issue #74, opened 2020-02-28
add land_surface_model attribute in model sfc history file


junwang-noaa:
Since model now can run with several different land surface model, POST needs some information in model output file to specify which land surface model is used. A land_surface_model attribute will be added to model sfc history files.

junwang-noaa:
The issue is resloved in PR #104. close issue.

PR #661, opened 2023-05-26
FV3 changes for refactored ozone physics scheme

dustinswales:
## Description
This PR contains modifications to accommodate changes to the NRL ozone physics parameterization. 

There are new interstitials variables for the data needed by the ozone scheme. The scheme data is ingested in GFS_typedefs:control_initialize() and provided to the scheme via interstitials, previously this data was stored on the HEAP within the scheme. Organizing the scheme this way removes the need for the scheme to access shared memory, making the scheme "stateless".

### Issue(s) addressed
[ccpp-physics #74](https://github.com/ufs-community/ccpp-physics/issues/74) 

## Testing
Coming soon...


ccpp/data/GFS_typedefs.F90
```diff
@@ -5370,28 +5388,51 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
-          oz_coeff = 4
-       else if (Model%oz_phys_2015) then
-          levozp   = 53
-          oz_coeff = 6
-       else
-          write(*,*) 'Logic error, ntoz>0 but no ozone physics selected'
-          stop
-       end if
+    !--- NRL ozone physics
+    if (Model%ntoz > 0) then
```
dustinswales, line 5413: @Qingfu-Liu Here is where read_o3data was ported to on the host side.

ccpp/data/GFS_typedefs.F90
```diff
@@ -5394,29 +5420,64 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
+       ! Get dimensions from data file
+       open(unit=Model%kozpl,file='global_o3prdlos.f77', form='unformatted', convert='big_endian')
```
grantfirl: I don't understand how this is working. `Model%kozpl` shouldn't be defined at this point. You've set the local `kozpl`, but haven't initialized the `Model` component yet, right? Same for `kozc`.

ccpp/data/GFS_typedefs.F90
```diff
@@ -5394,29 +5420,64 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
-          oz_coeff = 1
-       end if
+          Model%blatc   = blatc4
+       endif
+       Model%latsozp   = 2
+       Model%levozp    = 1
+       Model%timeoz    = 1
+       Model%oz_coeff  = 0
+       Model%dphiozc   = -(Model%blatc+Model%blatc)/(Model%latsozc-1)
     end if
```
grantfirl, line 5429: If we're moving the read_o3data code here, why not also the setindxoz code that also gets executed in the init phase? Since there are very similar 'setindxX` routines in ozinterp, h2ointerp, aerinterp, iccninterp, couldn't a common subroutine/function in here be used for all of them to reduce code replication?

ccpp/data/GFS_typedefs.F90
dustinswales: Great point. My guess is that it works with whatever garbage it's initialized to. I will fix this. 
Another thing, I don't think these file identifiers need to be interstitials anymore, as they aren't used in the physics, just in this module. I will also remove.

ccpp/data/GFS_typedefs.F90
dustinswales: Good point. I only moved over the data input from the _init step, not the other pieces.

dustinswales: @grantfirl 
A point worth mentioning. 
For a scheme to be stateless we can't allocate data on the HEAP during the physics initialization. To address this we can allocate the data on the host-side.

So in the case where a scheme has it's own external data, to achieve "statelessness", one could either a) allocate space here on the host side, and then read in during the physics initialization phase, or b) allocate and read in the data here. 
I chose the later to avoid having to open the file twice, and to also avoid passing the file identifier information into the physics. However, this raises the question you brought up "Why have any scheme initialization in the physics? Instead do it on the host?"

dustinswales:
@grantfirl I'm moving this PR, and the UWM PR, back to draft. It needs more input from othesr.

ccpp/data/GFS_typedefs.F90
dustinswales: outdated

ccpp/data/GFS_typedefs.meta
```diff
@@ -6651,6 +6657,18 @@
   units = flag
   dimensions = ()
   type = logical
+[levozp]
+  standard_name = number_of_levels_in_ozone_data
```
grantfirl: Why not reuse `vertical_dimension_of_ozone_forcing_data`?

grantfirl:
@dustinswales Can you pull in the latest ccpp-framework to this PR and update the description to say that we're doing this: https://github.com/NCAR/ccpp-framework/commit/219f2e9c88b7b774becac2bd1453696e105af1c4 ? I've tested this commit of the ccpp-framework using UFS-RTs and everything was bit-for-bit, so this should not change any baselines.

ccpp/data/GFS_typedefs.meta
dustinswales: Not sure why I did this. I will revert.

jkbk2004 (approved):

DusanJovic-NOAA (approved):

dustinswales:
@DusanJovic-NOAA Ready to merge

BrianCurtis-NOAA (approved):

PR #672, opened 2023-06-28
Export evaporative flux from ATM

DeniseWorthen:
## Description

Adds the evaporative flux to the available export fields from the ATM. 

Also adds a new procedure within the ``block_data_copy`` interface to prevent scaling of export values where the field contains a special value.

### Issue(s) addressed

- fixes #671 
- provides an incremental fix to resolution of CMEPS Issue https://github.com/NOAA-EMC/CMEPS/issues/93

## Testing

Testing results will be posted in the associated UFS PR https://github.com/ufs-community/ufs-weather-model/pull/1812

In brief, coupled tests using an active ATM and MOM6 will change baselines.

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)

- depends on https://github.com/NOAA-EMC/CMEPS/pull/96

Do PRs in upstream repositories need to be merged first?
If so add the "waiting for other repos" label and list the upstream PRs

- associated with https://github.com/ufs-community/ufs-weather-model/pull/1812

PR #675, opened 2023-06-30
Change signs of exported fields from ATM

### LINKED ISSUE #93: CMEPS is forced to use custom merges because sign of field imported from ATM is wrong or unit conversion is required
CMEPS has the capability to define generic merges between components. However, these generic merges are not available when the sign of the field needs to change, or the field needs to be converted to one of different units (ie, latent->evap). 

Because of this issue, multiple "custom" merges are required in ``med_phases_prep_ocn``.  For example, the sensible heat flux from the ATM has the wrong sign, requiring a custom merge:

```
       customwgt(:) = -ofrac(:)
       call med_merge_field(is_local%wrap%FBExp(compocn),      'Faxa_sen',  &
            FBinA=is_local%wrap%FBImp(compatm,compocn), fnameA='Faxa_sen', wgtA=customwgt, rc=rc)
       if (ChkErr(rc,__LINE__,u_FILE_u)) return
```

The ATM exports the latent heat flux, but the ocean requires the evaporative flux, which requires a custom merge to change the sign and units of the field:

```
       customwgt(:) = -ofrac(:) / const_lhvap
       call med_merge_field(is_local%wrap%FBExp(compocn),      'Faxa_evap', &
            FBinA=is_local%wrap%FBImp(compatm,compocn), fnameA='Faxa_lat' , wgtA=customwgt, rc=rc)
       if (ChkErr(rc,__LINE__,u_FILE_u)) return
```

The momemtum fluxes from the ATM also require a custom merge to change the sign. 

This adds needless complexity to maintain these custom merges when a better solution is to have the ATM export the coupling field with the right sign or unit. This also makes it possible to leverage the existing diagnostic capabilities with CMEPS to examine global budgets for UFS applications.

### LINKED ISSUE #669: ATM needs to export coupling fields to mediator with correct sign or unit
## Description
Currently specialized code is required in CMEPS because the export fields obtained from the ATM do not have the correct sign or unit. For example, MOM6 requires the evaporative flux, not the latent heat flux, and so specialized code is required in CMEPS just to change the sign and unit.  In addition, the momementum stresses and sensible heat flux also need a sign change within CMEPS. See https://github.com/NOAA-EMC/CMEPS/issues/93

## Solution
In ``atmos_model.F90``, both sign and units can be changed using the optional ``scale_factor`` argument. For example,

```
	    ! MEAN Zonal compt of momentum flux (N/m**2)
            case ('mean_zonal_moment_flx_atm')
              call block_data_copy(datar82d, GFS_data(nb)%coupling%dusfc_cpl, Atm_block, nb, scale_factor=-rtime, rc=localrc)
```

## Alternatives
Maintain existing code

## Testing:

I have a working [branch](https://github.com/DeniseWorthen/fv3atm/tree/feature/signs_units) which is B4B with current baselines with the exception of the mediator restart files (ie, the data is the same, but with an opposite sign).

However, this method does change the value of the "non-valid" value in the exported fields. Currently, these non-valid values are typically ``1.385E+34``.  An optional masking value could be used in the block_data interfaces to keep the values the same. I need to know where this value of ``1.385E+34`` is set however.

DeniseWorthen:
## Description

This PR changes the signs for the coupled fields exported to CMEPS by the ATM. These fields are the momentum fluxes, sensible and latent heat fluxes and the evaporative flux. 

### Issue(s) addressed

- fixes https://github.com/NOAA-EMC/CMEPS/issues/93
- fixes https://github.com/NOAA-EMC/fv3atm/issues/669

## Testing

Testing was carried out using https://github.com/ufs-community/ufs-weather-model/pull/1807. Full testing results are detailed there. In brief, in coupled and hafs tests, all baseline files are expected to be B4B with the exception of the mediator restart and history files. These files compare as different because the sign of the fields (but not the magnitude) has changed. All other tests reproduce. 

## Dependencies

- waiting on #672


DusanJovic-NOAA (approved):

junwang-noaa (approved):

FernandoAndrade-NOAA:
Testing for #[1807](https://github.com/ufs-community/ufs-weather-model/pull/1807) has finished successfully. Please continue with the merge process. @jkbk2004

PR #682, opened 2023-07-28
Updates for component land model

uturuncoglu:
## Description
This PR aims to add support for fully coupled atmosphere-land component configuration. 
- Along with this PR, UFS Weather Model will have capability to run land component in higher resolution than the atmosphere. 
- The PR introduces new input.nml option as `cpllnd2atm` to keep supporting side-by-side configuration (both component NoahMP and CCPP/NoahMP runs and component NoahMP does not send information to FV3, it is important particularly to compare the results coming from component land with the one that runs under CCPP/Physics). 
- The implementation also tested with C96 FV3ATM and C384 component land configuration. -
- There is no expected answer change for existing configurations.

### Issue(s) addressed
N/A

## Testing

The full regression test suite is run on Orion and all tests passed except land component related ones since new land component rely on newer version of ESMF (8.5.0) to read/write multi-tile netcdf files.

What compilers / HPCs was it tested with?  Intel / MSU's Orion

Following new RTs are added,
  - `control_p8_atmlnd`: fully coupled ATM-LND configuration
  - `control_restart_p8_atmlnd`: tests restart capability of fully coupled ATM-LND configuration
  - `datm_cdeps_lnd_era5`: component land model forced with ERA5
  - `datm_cdeps_lnd_era5_rst`: tests restart capability of component land model forced with 
 
Once UFS Weather model switches to new version of ESMF (8.5.0) there might have some answer changes due to some changes in ESMF interpolation routines. The full release note can be seen in the following link: https://github.com/esmf-org/esmf/releases/tag/v8.5.0

## Dependencies

Those branches should have matching names (ideally) N/A

- waiting on ufs-community/ccpp-physics/pull/94


ccpp/data/GFS_typedefs.F90
```diff
@@ -541,6 +541,20 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: dtsfcin_med(:)         => null()   !< sfc latent heat flux over ocean
     real (kind=kind_phys), pointer :: dqsfcin_med(:)         => null()   !< sfc sensible heat flux over ocean
     real (kind=kind_phys), pointer :: ulwsfcin_med(:)        => null()   !< sfc upward lw flux over ocean
+    !--- variables needed for cpllnd = .TRUE. and cpllnd2atm=.TRUE.
+    real (kind=kind_phys), pointer :: sncovr1_lnd(:)         => null()   !< sfc snow area fraction over land
+    real (kind=kind_phys), pointer :: qsurf_lnd(:)           => null()   !< sfc specific humidity
+    real (kind=kind_phys), pointer :: evap_lnd(:)            => null()   !< sfc latent heat flux over land
```
DeniseWorthen: Is this actually evap or latent? The comment and variable name disagree.

uturuncoglu: @DeniseWorthen Sorry for late reply. It is latent heat flux. I name it as `evap` to be consistent with NoahMP internal variable names. I know it is confusing but i could change it if you want.

DeniseWorthen: @uturuncoglu We've had this issue before w/ the latent/evap fields (exporting latent, but sending evap to ocean). Does it just need to be documented--for example :

```
real (kind=kind_phys), pointer :: evap_lnd(:)            => null()   !< sfc latent heat flux over land, converted to evaporative flux
```

uturuncoglu: @DeniseWorthen It is done. Thanks.

atmos_model.F90
```diff
@@ -2349,6 +2349,266 @@ subroutine assign_importdata(jdat, rc)
             endif
           endif
 
+! get surface snow area fraction: over land (if cpllnd=true and cpllnd2atm=true)
+!------------------------------------------------
+          fldname = 'mean_snow_area_fraction_lnd'
```
DeniseWorthen: Can you please update the import field names to ATM to either remove the "mean" or change them to "inst" here and in the cpl/module_cpl_fields. The fd_nems.yaml file will then need to be updated also. This will make your changes consistent with #704

uturuncoglu: @DeniseWorthen I think I need to wait first #704 is in. Right?

DeniseWorthen: We hope to get the series of PRs merged early next week which will include #704. This is to support the documentation effort. I think they are independent because they are new, added fields. But I wanted to be sure that when they are committed they're consistent.

atmos_model.F90
```diff
@@ -3072,12 +3332,6 @@ subroutine setup_exportdata(rc)
             ! bottom layer meridional wind (v)
             case('inst_merid_wind_height_lowest')
               call block_data_copy_or_fill(datar82d, DYCORE_data(nb)%coupling%v_bot, zeror8, Atm_block, nb, rc=localrc)
-            ! bottom layer zonal wind (u) from physics
-            case('inst_zonal_wind_height_lowest_from_phys')
-              call block_data_copy_or_fill(datar82d, GFS_data(nb)%Statein%ugrs, 1, zeror8, Atm_block, nb, rc=localrc)
```
junwang-noaa, line 3077: Do we know if these two fields are not used by any other components?

uturuncoglu: @junwang-noaa I add them in the past for side-by-side run since there was some difference in the wind components in that time but after discussed with @DeniseWorthen and test again, I see that  I could use the existing wind fields rather than introducing new fields. So, I removed them in this PR. I think no other components or configurations are using these fields.

atmos_model.F90
uturuncoglu: @DeniseWorthen I think this is already fixed. I am marking as resolved but let me know if you have any other changes.

junwang-noaa (approved):

atmos_model.F90
DeniseWorthen: yes, thanks for resolving all the aliases. It will make things simpler in the future.

DeniseWorthen (approved):

uturuncoglu:
Testing now.

uturuncoglu:
@haiqinli If you don't mind could you also test in your side. I have just merged your ccpp/physics and FV3 forks to this branch.

haiqinli:
Sure, you are welcome! - Haiqin

grantfirl:
https://github.com/ufs-community/ccpp-physics/pull/94 is merged. This is ready to update the ccpp/physics submodule pointer and revert .gitmodules.
@uturuncoglu

uturuncoglu:
@grantfirl okay. done it.

zach1221:
> @grantfirl okay. done it.
@uturuncoglu can you ensure the ccpp-physics hash is updated as well?
https://github.com/ufs-community/ccpp-physics/commit/d0c3294c7efa1531f489dcacff676b92acdab5a1

uturuncoglu:
@zach1221 let me check

uturuncoglu:
@zach1221 sorry, I think it is done.

zach1221:
@jkbk2004 this fv3atm sub-pr should be good to go. Can you please merge for us?

DeniseWorthen:
@jkbk2004 We need to ask ufuk to first merge the top of FV3, which contains documentation updates.

jkbk2004:
> @jkbk2004 We need to ask ufuk to first merge the top of FV3, which contains documentation updates.

@uturuncoglu there was a recent commit at fv3 develop. Can you sync up one more time?

uturuncoglu:
@jkbk2004 Okay, i synced. BTW, I still see changes in ccpp/physics.If you don't mind could you double check that all are fine.

jkbk2004:
> @jkbk2004 Okay, i synced. BTW, I still see changes in ccpp/physics.If you don't mind could you double check that all are fine.

I think changes just show diff against fv3/dev. Looks ok. I will let CI done and merge.

uturuncoglu:
@jkbk2004 Okay. That is great. BTW, thanks for all your help.

mkavulich:
@uturuncoglu Sorry to resurrect this discussion from the dead, but I had a question about your updates to the CCPP suite suite_FV3_GFS_v17_p8.xml: should the `sfc_land` scheme be added to other GFS v17 prototype suites as well? It seems like it only has an effect if `cpllnd2atm=.true.` in the namelist, so unless I'm missing something it couldn't hurt to add?

uturuncoglu:
@mkavulich I am not sure which suite file is used by GFS v17 but currently we have only in https://github.com/NOAA-EMC/fv3atm/blob/b3640007c51c8899533e1b835940a2030a768fcc/ccpp/suites/suite_FV3_GFS_v17_p8.xml#L46. That is right. You could add it to any suite file and it will update the land fluxes only `cpllnd2atm=.true.` case.

PR #704, opened 2023-09-27
mean to inst field names, part I

### LINKED ISSUE #1800: clarify "mean" and "inst" coupled fields exchanged
## Description

As a legacy from the NEMS mediator, many fields are exported or imported from the ATM with the name ``mean``. For example, ATM imports ``mean_ice_volume`` when in fact the field from CICE is an instantaneous field since the ATM and ICE are running concurrently in the fast loop. 

Even when coupling to ocean, the fields sent to the ocean from the ATM should be the instantaneous fields, since the mediator itself forms the average of these fields over the fast coupling loop in order to create the mean field to the ocean. Generally, these fields sent to the ocean are called ``mean`` fields, but they are the average of the ``inst`` fields from ATM or ICE (or merged between ATM and ICE). 

This issue arose while investigating the shortwave fields in issue #1767 

## Solution

Impose consistency in the fields which are imported or exported. For example, the ATM should export ``inst_sensi_heat_flx``, not ``mean_sensi_heat_flx`` for the ocean. For ATM exports, these two fields are not identical at every modelAdvance, but nearly so.

In the cases examined so far, changing to the instantaneous export fields shows very small (O~10-15) differences in the export fields in the first few hours after the initial time, after which differences start increasing as the model adjusts to the accumulated small differences.

Changing of ``mean_ice_volume`` to ``inst_ice_volume`` should not result in differences however, since in this case it is only the name of the field changing, and not the content.

In most cases, imposing this consistency on the ATM side will require only a modification to the fd_nems.yaml since both ``mean`` and ``inst`` fields are available the export field setup.

## Related to

- required by #1648

DeniseWorthen:
## Description

For fields imported to FV3atm from CICE6 and CMEPS, changes the field name alias used by FV3atm by removing the 'mean' prefix. Also renames the fields being exported by FV3atm as ``mean_prec_rate``, ``mean_fprec_rate`` and ``mean_prec_rate_conv`` to being ``inst`` fields in each case. The values of these fields used to fill the export state are unchanged. 

This is the first part of a two part commit sequence to change field names from mean->inst. This first commit will not change baselines. 

### Issue(s) addressed

- incrementally resolves ufs-community/ufs-weather-model/issues/1800

## Testing

Incremental testing has been done comparing this feature branch against a baseline created using the two dependent PRs on Hera. All baselines were B4B.

## Dependencies

- waiting on https://github.com/ufs-community/ufs-weather-model/pull/1894
- waiting on https://github.com/ufs-community/ufs-weather-model/pull/1906


DusanJovic-NOAA (approved):

atmos_model.F90
```diff
@@ -2015,7 +2015,7 @@ subroutine assign_importdata(jdat, rc)
 
 ! get upward LW flux:  for sea ice covered area
 !----------------------------------------------
-          fldname = 'mean_up_lw_flx_ice'
+          fldname = 'lwup_flx_ice'
```
junwang-noaa, line 2018: I assume, the names without "mean" means inst fields?

junwang-noaa (approved):

atmos_model.F90
DeniseWorthen: I could have picked a better alias, but yes. In general, I think the fields exchanged between components should simply being understood as the field value at the end of the model advance, the "instantaneous" value.

zach1221:
@DeniseWorthen testing is finished on ufs-wm PR [#1925](https://github.com/ufs-community/ufs-weather-model/pull/1925).  @jkbk2004 we have the approvals needed, can you please go ahead to merge this PR?

PR #695, opened 2023-09-08
Quartet of bug fixes for: c3 scheme, quilting restart with 32-bit physics, and string length mismatch in dycore plus PR #705 and #699

SamuelTrahanNOAA:
## Description

Bug fixes:

1. *The write component no longer calls recover_fields for restart files.* (From @DusanJovic-NOAA.) Doing so causes sporadic crashes due to corrupted longitudes when 32-bit physics is used. This only happens for restart files, who don't need recover_fields anyway. Hence, this PR restricts that call to history files.
2. *Correct a string length mismatch in the dycore.* This causes an abort when the gnu compiler is run with debug flags turned on. One of the checks enabled is to abort on standard violations of this type.
3. *Changes some subroutine array argument declarations so the c3 scheme can handle null pointers.* This is required by the CCPP coding standards anyway, for that very reason. Without this change, the FV3_HRRR_c3 suite crashes with the gnu compiler.
4. *Initialize arrays in dycore.* (From @DusanJovic-NOAA.) Two arrays in the dycore were not initialized. That caused crashes in the quilting restart when 32-bit physics is used.

Included in this PR are:

- #705 
- #699 
- https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/276
- https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/285

### Issue(s) addressed

- fixes https://github.com/NOAA-EMC/fv3atm/issues/694
- fixes https://github.com/ufs-community/ccpp-physics/issues/112

#### How were these changes tested?  
New or improved ufs-weather-model regression tests

#### What compilers / HPCs was it tested with?  

jet, hera, cheyenne, hercules, orion, and cactus (wcoss2)

#### Are the changes covered by regression tests?

Some ufs-weather-model regression tests are added or modified.

#### Have the ufs-weather-model regression test been run? On what platform?  

jet, hera, cheyenne, hercules, orion, and cactus (wcoss2)

#### Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.

Yes. All tests with FV3 will change due to https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/276

#### Please commit the regression test log files in your ufs-weather-model branch

Will do.

## Dependencies

- waiting on https://github.com/ufs-community/ccpp-physics/pull/102
- waiting on https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/280

PR #705, opened 2023-09-27
Remove nine 3D arrays from CLM Lake model

### LINKED ISSUE #112: CLM Lake restarts nine 3D fields that are constant
# Description

These fields in the CLM Lake model are constant:

- lake_z3d
- lake_dz3d
- lake_soil_watsat3d
- lake_csol3d
- lake_soil_tkmg3d
- lake_soil_tkdry3d
- lake_soil_tksatu3d
- lake_clay3d
- lake_sand3d

The later seven are 2D arrays that are needlessly duplicated in the vertical (lake k) dimension. The first two are 3D. All of these are sparse arrays, zero everywhere except lake points.

All nine should be removed.

## Steps to Reproduce

1. Read the code. The lakeini in clm_lake.f90 calculates those variables. They're never modified after that.

## Additional Context

The CLM Lake Model is taking up too much disk space in the RRFS. Removing these nine variables would reduce that problem.

Problem identified by @hu5970 and @tanyasmirnova

SamuelTrahanNOAA:
## Description

There are nine 3D arrays in CLM Lake that are constant. Seven of them are 2D arrays copied to all vertical levels. All are cheap to calculate.

This PR removes all nine arrays. When the numbers are needed, they're recalculated.

To do this, I had to add a 2D array: the lakedepth before correction.

### Issue(s) addressed

- fixes https://github.com/ufs-community/ccpp-physics/issues/112

A 3km RRFS case was run, removing one array at a time until all nine were gone.

Hera intel & gnu.

Existing tests are sufficient.

Hera

Yes.

1. Any configuration running clm lake that verifies restart files will change because of removing nine variables and adding two.
2. Removing kind_phys arrays means some calculations are promoted to kind_lake (double precision). This causes 32-bit physics cases with lake data running clm lake to change results.
3. When run with optimization, some clm lake cases with lake data will change results due to a different order of operations.

Done.

## Dependencies

- waiting on https://github.com/ufs-community/ccpp-physics/pull/113

### DIFF
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -431,13 +431,7 @@ module GFS_typedefs
 
     ! CLM Lake model internal variables:
     real (kind=kind_phys), pointer :: lake_albedo(:)     => null()  !
-    real (kind=kind_phys), pointer :: lake_z3d(:,:)     => null()  !
-    real (kind=kind_phys), pointer :: lake_dz3d(:,:)    => null()  !
-    real (kind=kind_phys), pointer :: lake_soil_watsat3d(:,:) => null()  !
-    real (kind=kind_phys), pointer :: lake_csol3d(:,:)   => null()  !
-    real (kind=kind_phys), pointer :: lake_soil_tkmg3d(:,:)   => null()  !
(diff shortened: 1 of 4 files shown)


DusanJovic-NOAA:
Merged via #695

PR #743, opened 2023-12-20
Use MPI_Bcast instead of multiple p2p messages to update nest from parent

dkokron:
Performance profiling of a HAFS case on NOAA systems revealed significant of time was spent in fill_nested_grid_cpl(). The fill_nested_grid_cpl() routine from FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90 is showing up as a performance bottleneck. This routine gathers a global SST field (6,336,000 bytes) onto rank 0, then broadcasts that field to all ranks in the nest. The code uses point-to-point (p2p) messages (Isend/Recv) from rank 0 to the nest ranks. This communication pattern is maxing out the SlingShot-10 link on the first node resulting in a .15s hit every fifth time step.

The proposed fix is to modify the relevant FV3 code to use a single MPI_Bcast (via mpp_broadcast()) instead of multiple point-to-point messages.  The use of mpp_broadcast depends on a fix to FMS that was merged on 16 June and is available in version 2023.02 of that package.
https://github.com/NOAA-GFDL/FMS/pull/1246

This PR depends on merging of PR 272 into the GFDL_atmos_cubed_sphere
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/272

I ran the UFS regression suite on acorn and cactus. Both runs resulted in "REGRESSION TEST WAS SUCCESSFUL"

This change is zero-diff.  No need to update baselines


BrianCurtis-NOAA:
Waiting on ACS approval and hash update and .gitmodules revert.

jkbk2004 (approved):

BrianCurtis-NOAA (approved):

PR #744, opened 2023-12-22
Add option to write 3d soilt, soilw and soill variables to a history file

### LINKED ISSUE #1055: Add vertical dimension to certain surface fields
## Description
This discussion started in https://github.com/ufs-community/ufs-weather-model/issues/1052 but continuing it here.

Many of the fields in the `sfcfNNN.nc` files look like so:
```
	float soilt1(time, tile, grid_yt, grid_xt) ;
	float soilt2(time, tile, grid_yt, grid_xt) ;
	float soilt3(time, tile, grid_yt, grid_xt) ;
	float soilt4(time, tile, grid_yt, grid_xt) ;
```
Which is 4 layers of soil temperature. 

I think it would make more sense for these fields to be instead like so:
```
	float soilt(time, tile, zsoil, grid_yt, grid_xt) ;
```
or something like that.

This would greatly simplify I/O routines as it would be a few variables with a vertical dimension rather than k*variables where k is the number of layers. I also suspect that the total file size would decrease with fewer variables being defined (but not sure by how much).

## Solution
Update fields that should have a vertical component to include one in the write-grid component output.

## Alternatives
Keep the files as is.

## Related to
The UPP and other downstream users may depend on this change?

DusanJovic-NOAA:
## Description

This PR adds three 3D soil variables (soilt, soilw and soill) to a history file if corresponding entries are specified in diag_table file. The new 3d versions of these variables is identical to a set of 2d variables currently available (soilt1, soilt2, etc)

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/1055

## Testing

How were these changes tested?  Yes
What compilers / HPCs was it tested with?  Intel and GNU / Hera
Have the ufs-weather-model regression test been run? Yes.
On what platform? Hera  
- Will the code updates change regression test baseline? Yes, control_CubedSphereGrid_parallel regression test is modified to output 3d version of soilt, soilw and soill variables.

## Dependencies

N/A


BrianCurtis-NOAA (approved):

FernandoAndrade-NOAA:
@jkbk2004 could you please approve this when available, thanks! Testing on #[2062 ](https://github.com/ufs-community/ufs-weather-model/pull/2062) has completed successfully.

jkbk2004 (approved):

Issue #729, opened 2023-12-07
Add CI build & unit testing


AlexanderRichert-NOAA:
## Description
Creating this issue to discuss and plan the addition of CI workflows to the repository for building and testing all code modifications.

## Solution
I've added a GCC-and-MPICH-based CI workflow in my fork; currently it only builds, I have not added any tests: https://github.com/AlexanderRichert-NOAA/fv3atm/tree/develop
Here's my first successful run of the CI, which currently has two jobs, for building with 32 and 64 bit reals: https://github.com/AlexanderRichert-NOAA/fv3atm/actions/runs/7122152262

## Next steps
If the CI so far looks good, then the main next step is to add unit tests with CTest. I would recommend we add a checklist to this issue to establish at least two or three initial tests, merge it into develop, then create one or more issues for adding further tests. Other possible additions we can discuss include the use of Intel, other MPI flavors, code coverage checks, and documentation (e.g., Doxygen).

AlexanderRichert-NOAA:
@edwardhartnett @junwang-noaa

edwardhartnett:
This is great progress. Can you put up a PR with the changes needed and the workflows?

AlexanderRichert-NOAA:
See #730

edwardhartnett:
Should this issue be closed? Note #906 .

DusanJovic-NOAA:
Fixed in #752 merged via #906

PR #752, opened 2024-01-13
Add unit testing

AlexanderRichert-NOAA:
## Description
This PR:
 - Adds unit testing with the CTest framework, plus code coverage checking with gcovr; and
 - adds a proposed update to the PR template that has checkboxes for contributors to certify that they are appropriately adding unit tests and documentation to the code.

I relocated the build/ directory in the CI to be under the fv3atm repo directory because it made the gcovr setup simpler. I've also added some logic that ensures that the gcovr report and doxygen output only get generated for one of the jobs, but still under GCC.yml to keep things consolidated.

### Issue(s) addressed
Addresses #729

## Testing
Tested on personal computer. All CI jobs successful.

## Dependencies
none


AlexanderRichert-NOAA:
@junwang-noaa @edwardhartnett @DusanJovic-NOAA it looks like I can't add reviewers, but please take a look at this PR, I believe it's good to go.

ci/CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ endif()
 find_package(bacio 2.4.0 REQUIRED)
 find_package(sp 2.3.3 REQUIRED)
 find_package(w3emc 2.9.2 REQUIRED)
+find_package(PIO REQUIRED)
```
DusanJovic-NOAA: Why is PIO required?

tests/test_fv3_io_def.F90
```diff
@@ -0,0 +1,130 @@
...
+! Alex Richert, 11 Jan 2024
+
+program test_fv3_io_def
+
+  use module_fv3_io_def
+
+  implicit none
+
+  num_pes_fcst = 2
+  if (num_pes_fcst .ne. 2) stop 1
```
DusanJovic-NOAA: Which function (code) in fv3atm is this unit test testing? As far as I can see num_pes_fcst will always be equal to 2, and this test will always pass, regardless on what is actually changed in the fv3atm code. I do not see the purpose of this test.

AlexanderRichert-NOAA: None. I added it initially as an easy way to test the code, since it relies on the variables being defined in module_fv3_io_def (so it would catch an instance of, say, someone accidentally removing a variable in there).

ci/CMakeLists.txt
AlexanderRichert-NOAA: I had added a test for module_write_netcdf that was failing because pio wasn't linked, and this took care of it. I don't think it's critical for this PR but I believe it'll be needed for future tests.

AlexanderRichert-NOAA: It only just occurred to me that this may be an issue with FindESMF.cmake, i.e., it may not be passing the right linker flags. I'll look into it further.

tests/test_fv3_io_def.F90
DusanJovic-NOAA: If someone accidentally removes a variable from module_fv3_io_def the compilation will fail. I do not see the need for a test that only assigns the value to a variable and does nothing with a that variable, only to make sure that variable is not removed.

AlexanderRichert-NOAA: Yeah I'm fine removing the test, I don't think I need the easy proof of concept with the other test that's there.

ci/CMakeLists.txt
AlexanderRichert-NOAA: It's not necessary currently, so I'm removing it for now and I can circle back to it if it becomes an issue for future tests.

DusanJovic-NOAA (approved):

AlexanderRichert-NOAA:
@junwang-noaa please review. I suggest squash & merge.

junwang-noaa (approved):

junwang-noaa:
@DusanJovic-NOAA Do we have an fv3 PR that will be committed in the upcoming ufs wm commites? Can we coordinate it with this one (ask the person to merge fv3atm when this PR is committed)? Thanks

DusanJovic-NOAA:
I see #767 scheduled for merge in the ufs-weather-model commit queue. @grantfirl can you please merge this branch with your #767 PR. This PR updates github workflow to run unit tests.

AlexanderRichert-NOAA:
@DusanJovic-NOAA I've merged in the current develop branch so it includes your changes. Otherwise everything should basically be as it was before.

climbfuji:
Pulled into https://github.com/NOAA-EMC/fv3atm/pull/798

DusanJovic-NOAA:
Merged via #798

Issue #289, opened 2021-04-20
Shall I add doxygen to the build system?


edwardhartnett:
Shall I add doxygen to the build system, which would allow programmers to start using doxygen in the code?

junwang-noaa:
Yes, it is good to have it in the build system.

edwardhartnett:
This is being held back by the lack of self-build capability in fv3atm. We can't do a doxygen build in CI if fv3atm cannot even be compiled successfully in CI.

edwardhartnett:
I will close this issue until there is greater interest, and a fixed build system.

edwardhartnett:
OK, interest in testing fv3atm has led to interest in documentation too.

The problems start because of the way we're lumping so many projects together using git submodules, rather than releasing them as separate libraries and then linking them in the usual way.

For example, the doc build of fv3atm will not work because both ccpp and upp include a "doc" target in their cmake files.

```
CMake Error at upp/docs/CMakeLists.txt:10 (add_custom_target):
  add_custom_target cannot create target "doc" because another target with
  the same name already exists.  The existing target is a custom target
  created in source directory
  "/home/runner/work/fv3atm/fv3atm/fv3atm/ccpp/framework/doc".  See
  documentation for policy CMP0002 for more details.


```
See https://github.com/edwardhartnett/fv3atm/actions/runs/7371629648/job/20059198642?pr=2

PR #757, opened 2024-01-16
doxygen changes for io/module_fv3_io_def.F90 and module_fv3_config.F90

edwardhartnett:
## Description

Doxygen changes.

### Issue(s) addressed

Part of #289

## Testing

Documentation changes only.

## Dependencies

N/A


io/module_fv3_io_def.F90
```diff
@@ -1,36 +1,126 @@
...
+  !> ???  
+  integer           :: nbdlphys 
+
+  !> ???  
+  integer           :: iau_offset 
+
+  !> ???  
+  logical           :: lflname_fulltime 
+
+  !> ???
```
DeniseWorthen: So the next step is to, for example, add ?

```
integer :: num_pes_fcst.            !> Number of processors used in forecast run
integer :: wrttasks_per_group   !> Number of write tasks per group
integer :: write_groups               !> Number of write groups
....
```

These are local variables though, right? I think when I did this in UFS_UTILS, I did not add doxygen for local variables, only argument variables.

edwardhartnett: Yes, the next step is to add info wherever I have put "???". But when documenting with doxygen we can do either:

`integer :: i !< i is an index
`

or we can do:

```
!> i is an index
integer :: i

```
Note that in one case we use >, in the other we use <.

These are module variables and need to be documented.

Local variables in subprograms do not need to be documented with doxygen.

DeniseWorthen: Can we define a style that we will use? I much prefer the one-line version because it keeps the code more compact. 

Thanks for clarifying the local/module. I didn't look carefully enough at the diff.

edwardhartnett: We can use the one-line version where the documentation will fit on the rest of the line, but frequently it will be longer than that.

We are using a style which matches the way we do documentation in the rest of the EMC software, including UFS_UTILS.

edwardhartnett: BTW the way to add documentation is not to add it in code review, but to do a PR against my branch, so I don't have to cut and paste all the documentation. ;-)

DeniseWorthen: I agree that having to wrap w/ to another line (using !< and then !!) can be a problem when doing, for example, auto-indenting but I still prefer for both compactness and readability.

edwardhartnett: Each project, and each coder, can use variations of documentation style, as long as they do not stray too far from the way we doing things in other EMC projects.

So if you want to document like this:

```
integer :: i !< some long documentation about i
!! which continues on the next line.

```
Instead of:

```
!> some long documentation about i
!! which continues on the next line.
integer :: i 

```
Then I would be happy to get the good documentation and would not consider the formatting difference to be important.

io/module_fv3_io_def.F90
```diff
@@ -9,91 +9,91 @@ module module_fv3_io_def
   use esmf, only     : esmf_maxstr
   implicit none
 
-  !> Number of processors used in forecast run.
+  !> Number of processors used in the forecast grid component
```
DeniseWorthen, line 12: Should we consider adding a short paragraph at the top of each module to describe what the module is doing?  There isn't much context for some of the various parameter definitions. You can say the variable ``write_groups`` is the number of write_groups, but that doesn't help the user understand how a write_group number is used.

edwardhartnett: Yes absolutely! More detail is very welcome.

junwang-noaa (approved):

junwang-noaa:
I tested the branch in UFS wm, ful RT passed on hera.
/scratch1/NCEPDEV/nems/Jun.Wang/nems/vlab/20240118/ufs-weather-model/tests/logs/RegressionTests_hera.log.

DusanJovic-NOAA (approved):

PR #767, opened 2024-01-18
Fix -check all runtime errors (for CCPP) using zero-size allocations

grantfirl:
## Description

This PR at least partially addresses https://github.com/ufs-community/ufs-weather-model/issues/2023 (at least any runtime errors associated with the CCPP). The strategy is to simply make sure that every variable is allocated, although for variables that had been conditionally-allocated to save memory, they are now being allocated with zero size. This has a ripple effect in other repos where one needs to change from checking whether a variable is allocated to whether the size is > 0.

Note: There was also some cleanup done to remove variables that are declared in GFS_ and CCPP_typedefs.F90 but are not being used anywhere anymore. In addition, the ACTIVE attribute in the metadata files was double-checked for every conditionally-allocated variable.

### Issue(s) addressed

https://github.com/ufs-community/ufs-weather-model/issues/2023

## Testing

This was tested with the full RT suite on Hera with the `-check all` compilation flag added in the ufs-weather-model Intel.cmake file. All tests completed successfully with no BL changes.

## Dependencies

- waiting on https://github.com/ufs-community/ccpp-physics/pull/159
- waiting on a PR for https://github.com/climbfuji/ccpp-framework/tree/feature/dummy_allocate_inactive
- waiting on https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/315


grantfirl:
This is replaced by https://github.com/NOAA-EMC/fv3atm/pull/789 that is targeted at the RRFS release branch. This will likely NOT need to go to develop since a longer-term fix is forthcoming.

PR #805, opened 2024-03-20
Update github workflow

### LINKED ISSUE #804: GitHub workflow run fails
## Description
The gihub workflow run fails with "Cannot open module file ‘mpi_f08.mod’ for reading" error. See #770

### To Reproduce:
What compilers/machines are you seeing this with? Github workflow run
Give explicit steps to reproduce the behavior.
1. Look at any recent workflow log file

## Additional context
In addition to fixing this issue we should also remove GNU compiler flag '-fallow-argument-mismatch'. Not needed anymore.

## Output
```
/home/runner/work/fv3atm/fv3atm/fv3atm/ccpp/framework/src/ccpp_types.F90:22:9:

   22 |     use mpi_f08, only: MPI_Comm
      |         1
Fatal Error: Cannot open module file ‘mpi_f08.mod’ for reading at (1): No such file or directory
compilation terminated.
```

DusanJovic-NOAA:
## Description

- Updated the gcc compiler version from 11 to 12.
- Add 'spack clean --all' to reduce the size of the github cache file, Currently cache file is about 3.5 GB, after this change cache file is about 150 - 200 MB.
- Split GCC workflow into two steps, the first step only builds spack, the second step builds the fv3atm. Currently spack stack is built for each combination of compile/mpi/cmake_flags even though spack does not depend on cmake flags (those are fv3atm specific flags).

### Issue(s) addressed

- fixes #804

## Testing

How were these changes tested?  GitHub CI

## Dependencies

N/A

### DIFF
--- a/.github/workflows/GCC.yml
+++ b/.github/workflows/GCC.yml
@@ -58,6 +58,7 @@ jobs:
         spack config add "packages:mpi:require:'${{ matrix.mpi }}'"
         spack concretize |& tee ${SPACK_ENV}/log.concretize
         spack install -j2 --fail-fast
+        spack clean --all
 
     - name: build-fv3atm
       run: |
--- a/.github/workflows/GCC.yml
+++ b/.github/workflows/GCC.yml
@@ -26,14 +26,14 @@ jobs:
     steps:
 
     - name: checkout-fv3atm
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         path: ${{ github.workspace }}/fv3atm
         submodules: recursive
 
     - name: cache-spack
       id: cache-spack
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}/spack-develop
         key: spack-${{ hashFiles('fv3atm/ci/spack.yaml') }}-gcc${{ matrix.gcc_ver }}-${{ matrix.mpi }}
@@ -72,14 +72,14 @@ jobs:
         sudo apt-get install doxygen graphviz
 
     - name: checkout-fv3atm
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         path: ${{ github.workspace }}/fv3atm
         submodules: recursive
 
     - name: cache-spack
       id: cache-spack
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}/spack-develop
         key: spack-${{ hashFiles('fv3atm/ci/spack.yaml') }}-gcc${{ matrix.gcc_ver }}-${{ matrix.mpi }}
@@ -102,16 +102,16 @@ jobs:
         make -j2
         ls -l /home/runner/work/fv3atm/fv3atm/fv3atm/io
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
-        name: docs
+        name: docs-gcc${{ matrix.gcc_ver }}-${{ matrix.mpi }}-${{ matrix.cmake_opts }}
         path: |
           build/docs/html
 
     - name: debug-artifacts
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       if: ${{ failure() }}
       with:
-        name: ccpp_prebuild_logs
+        name: ccpp_prebuild_logs-gcc${{ matrix.gcc_ver }}-${{ matrix.mpi }}-${{ matrix.cmake_opts }}
         path: ${{ github.workspace }}/build/ccpp/ccpp_prebuild.*
 
--- a/.github/workflows/GCC.yml
+++ b/.github/workflows/GCC.yml
@@ -31,6 +31,12 @@ jobs:
         path: ${{ github.workspace }}/fv3atm
         submodules: recursive
 
+    - name: install-cmake
+      run: | 
+        cd ${{ github.workspace }}
+        curl -f -s -S -R -L https://github.com/Kitware/CMake/releases/download/v3.29.2/cmake-3.29.2-Linux-x86_64.tar.gz | tar -zx
+        echo "${{ github.workspace }}/cmake-3.29.2-linux-x86_64/bin" >> $GITHUB_PATH
+
     - name: cache-spack
       id: cache-spack
       uses: actions/cache@v4
@@ -70,6 +76,12 @@ jobs:
     - name: install-doxygen
       run: |
         sudo apt-get install doxygen graphviz
+
+    - name: install-cmake
+      run: | 
+        cd ${{ github.workspace }}
+        curl -f -s -S -R -L https://github.com/Kitware/CMake/releases/download/v3.29.2/cmake-3.29.2-Linux-x86_64.tar.gz | tar -zx
+        echo "${{ github.workspace }}/cmake-3.29.2-linux-x86_64/bin" >> $GITHUB_PATH
 
     - name: checkout-fv3atm
       uses: actions/checkout@v4
--- a/.github/workflows/GCC.yml
+++ b/.github/workflows/GCC.yml
@@ -98,6 +98,7 @@ jobs:
         export CC=mpicc
         export CXX=mpicxx
         export FC=mpif90
+        cat /home/runner/work/fv3atm/fv3atm/spack-develop/opt/spack/linux-ubuntu22.04-zen2/gcc-12.3.0/fms-2023.04-*/lib/cmake/fms/fms-config.cmake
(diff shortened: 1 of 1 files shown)


AlexanderRichert-NOAA (approved):
This all looks good to me. Just out of curiosity what's the motivation for going from gcc 11 to 12?

BrianCurtis-NOAA (approved):

DusanJovic-NOAA:
> This all looks good to me. Just out of curiosity what's the motivation for going from gcc 11 to 12?

If I remember correctly gcc 11 had some issues with mpich.

DusanJovic-NOAA:
I see this warning:

Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/cache@v3, actions/upload-artifact@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

here: https://github.com/NOAA-EMC/fv3atm/actions/runs/8375122977

AlexanderRichert-NOAA:
All of those actions are currently at version 4.x, so changing `@v3` to `@v4` should make the warnings go away.

DusanJovic-NOAA:
This PR will probably create conflicts with #752. @AlexanderRichert-NOAA I do not know which PR we'll be merging first but feel free to combine the changes from this PR with your 'Add unit testing' PR.

AlexanderRichert-NOAA:
@DusanJovic-NOAA I'll take a look, thanks

.github/workflows/GCC.yml
```diff
@@ -58,6 +53,36 @@ jobs:
...
+        sudo apt-get install doxygen graphviz
+
+    - name: checkout-fv3atm
+      uses: actions/checkout@v4
+      with:
+        path: ${{ github.workspace }}/fv3atm
+        submodules: recursive
+
+    - name: cache-spack
+      id: cache-spack
```
AlexanderRichert-NOAA, line 93: Since the Spack cache isn't/shouldn't be modified in the part of the workflow, and since the cache should definitely exist at this point, I think we could use `uses: actions/cache/restore@v4` (https://github.com/actions/cache).

DusanJovic-NOAA: action/cache/restore works for build_fv3atm job. Thanks for suggestion.

junwang-noaa (approved):

BrianCurtis-NOAA (approved):

DusanJovic-NOAA:
@AlexanderRichert-NOAA did you have a chance to look into potential conflicts in your #752 PR? Or should we merge this PR now as it is?

AlexanderRichert-NOAA:
Yes, thanks. I don't foresee any problem with merging this first.

DusanJovic-NOAA:
Unfortunately workflow is failing, this time with:
```
CMake Error at /home/runner/work/fv3atm/fv3atm/spack-develop/opt/spack/linux-ubuntu22.04-zen2/gcc-12.3.0/fms-2023.04-cmrjkxgqmcz74sjpou2qtdgwag7mavuq/lib/cmake/fms/fms-config.cmake:11 (message):
  File or directory referenced by variable FMS_INSTALL_PREFIX does not exist
  !
Call Stack (most recent call first):
  /home/runner/work/fv3atm/fv3atm/spack-develop/opt/spack/linux-ubuntu22.04-zen2/gcc-12.3.0/fms-2023.04-cmrjkxgqmcz74sjpou2qtdgwag7mavuq/lib/cmake/fms/fms-config.cmake:52 (set_and_check)
  ci/CMakeLists.txt:58 (find_package)
  CMakeLists.txt:18 (include)
```

Previous run from this branch finished successfully: https://github.com/NOAA-EMC/fv3atm/actions/runs/8375122977

DusanJovic-NOAA:
I tried to reproduce the build failure in the github workflow by building spack stack on my desktop computer but I couldn't. It's a different OS (Red Hat vs. Ubuntu), slightly different compiler and mpi versions. Building the spack stack failed first while building the ip library. In CI ip@develop is specified which needs LAPACK, which I do not have installed. Model currently uses ip/4.3.0 not ip/develop. I also noticed that the FMS version compiled in the workflow stack is 2023.04 while the model currently uses 2023.02.01. The versions of ESMF are also different 8.4.3 in CI vs. 8.5.0 that model currently needs. It's strange that in CI some libraries are newer that what model currently uses while some others are older.

We should still try to use the same versions everywhere, just for consistency. 

However, none of these explains why compilation fails in the github workflow.

AlexanderRichert-NOAA:
How recently was the cache rebuilt? That's the only thing I can think of... Can you try printing the value of that variable that it's complaining about? Or if nothing else, maybe `cat /home/runner/work/fv3atm/fv3atm/spack-develop/opt/spack/linux-ubuntu22.04-zen2/gcc-12.3.0/fms-2023.04-*/lib/cmake/fms/fms-config.cmake` and see if that yields any clues. Let me know if you'd like me to dive into this further.

DusanJovic-NOAA:
When I look at the fms-config.cmake in the install tree on my desktop, which should be either identical or very similar to the one in the github cache, I see:

```
$ more ./lib/cmake/fms/fms-config.cmake

####### Expanded from @PACKAGE_INIT@ by configure_package_config_file() #######
####### Any changes to this file will be overwritten by the next CMake run ####
####### The input file was FMSConfig.cmake.in                            ########

get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)

macro(set_and_check _var _file)
  set(${_var} "${_file}")
  if(NOT EXISTS "${_file}")
    message(FATAL_ERROR "File or directory ${_file} referenced by variable ${_var} does not exist !")
  endif()
endmacro()

. . .


set(FMSVersion "${PACKAGE_VERSION}")
set_and_check(FMS_INSTALL_PREFIX "${PACKAGE_PREFIX_DIR}")

```

Variable `FMS_INSTALL_PREFIX` is set to be equal to `PACKAGE_PREFIX_DIR`, which is set to `${CMAKE_CURRENT_LIST_DIR}/../../../`

`CMAKE_CURRENT_LIST_DIR` should be `.../fms-2023.04-*/lib/cmake/fms` and `PACKAGE_PREFIX_DIR` (and `FMS_INSTALL_PREFIX`) should then be `.../fms-2023.04-*`

I do not understand how this file (fms-config.cmake) exists but it's parent does not exist.

DusanJovic-NOAA:
I added `cat ...../fms-config.cmake` to github workflow file and I see:
```
####### Expanded from @PACKAGE_INIT@ by configure_package_config_file() #######
####### Any changes to this file will be overwritten by the next CMake run ####
####### The input file was FMSConfig.cmake.in                            ########

get_filename_component(PACKAGE_${CMAKE_FIND_PACKAGE_NAME}_COUNTER_1 "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)

....
```
which is different than in the equivalent fms-config.cmake file on hera, for example:

```
$ head /scratch1/NCEPDEV/nems/role.epic/spack-stack/spack-stack-1.6.0/envs/unified-env-rocky8/install/intel/2021.5.0/fms-2023.04-ywa3zzf/lib/cmake/fms/fms-config.cmake 

####### Expanded from @PACKAGE_INIT@ by configure_package_config_file() #######
####### Any changes to this file will be overwritten by the next CMake run ####
####### The input file was FMSConfig.cmake.in                            ########

get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)

macro(set_and_check _var _file)
  set(${_var} "${_file}")
  if(NOT EXISTS "${_file}")

```

That first line should set `PACKAGE_PREFIX_DIR`, which is later used to set `FMS_INSTALL_PREFIX`.

I do not know why this is different in workflow cache than on Hera. Could it be FMS issue, or spack issue, or cmake issue, I'm not sure.

DusanJovic-NOAA:
Looks like this is an issue with the specific version of cmake (3.29.1) currently used by github workflow:

https://gitlab.kitware.com/cmake/cmake/-/issues/25827
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9420

https://github.com/microsoft/vcpkg/issues/37968

AlexanderRichert-NOAA:
Hm that's unfortunate, good find though. Unless there's another version of cmake available in the runners, like under /usr/local somewhere, then maybe the way to go is to build it through Spack, which takes a few minutes, but at least it'll get cached. I think that would only require removing cmake from the `spack external find` call.

DusanJovic-NOAA:
Or we can just wait for github to update their VMs with the latest version, which I hope they'll do soon and we can then just remove the step that installs cmake from a tar file. Three weeks ago this worked just fine, which means in the meantime version 3.29.1 has been released and github picked it. We should also add status message to our CMakeLists.txt to print the version of cmake used, which is useful in situations like this.

DusanJovic-NOAA:
Merged via #811

Issue #760, opened 2024-01-17
Update doxygen documentation in fv3atm so that doxygen builds without warnings


edwardhartnett:
## Description
As discussed at a recent meeting, we will update the doxygen documentation in fv3atm. We want to build doxygen with all warnings turned on, and have the fv3atm documentation build without warning. When this is the case, we can turn on documentation checking in CI, and no one will be able to merge code without full documentation.

## Solution

We have to add missing documentation elements everywhere. Programmers must fill in missing information. This will take a while!

@AlexanderRichert-NOAA and @edwardhartnett will add the missing doxygen elements, but only fv3atm programmers can provide the needed information.

## Alternatives

Doxygen is what we use all across the  EMC, so we don't want fv3atm to be different from all other EMC software. Also doxygen is effective, simple, and provides all the features we need. It's also very common and well-understood by the programming community.

## Testing:

The CI build has been modified to build docs and upload them to GitHub, so anyone can download a zip file from GitHub actions and see the built documentation.

## Dependent PRs:

N/A.

PR #811, opened 2024-03-28
Update upp submodule

WenMeng-NOAA:
## Description

This PR is for updating upp submodule with recent changes from UPP repository which include Rocky8 transition on Hera and GFS HR4 support.
What bug does it fix, or what feature does it add?  
Is a change of answers expected from this PR?  

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
- fixes #809

## Testing

How were these changes tested?  
What compilers / HPCs was it tested with?  
Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)  
Have the ufs-weather-model regression test been run? On what platform?  
- Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.
- Please commit the regression test log files in your ufs-weather-model branch

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)

Do PRs in upstream repositories need to be merged first?
If so add the "waiting for other repos" label and list the upstream PRs

PR #819, opened 2024-04-10
doxygenated fv3_cap.F90

AlysonStahl-NOAA:
## Description

More doxygen

### Issue(s) addressed

Part of #760 

## Testing

Documentation changes only

## Dependencies

N/A

### DIFF
--- a/fv3_cap.F90
+++ b/fv3_cap.F90
@@ -1,16 +1,20 @@
-!--------------- FV3 ATM solo model ----------------
-!
-!*** The FV3 atmosphere grid component nuopc cap
-!
-! Author:  Jun Wang@noaa.gov
-!
-! revision history
-! 11 Oct 2016: J. Wang          Initial code
-! 18 Apr 2017: J. Wang          set up fcst grid component and write grid components
-! 24 Jul 2017: J. Wang          initialization and time stepping changes for coupling
-! 02 Nov 2017: J. Wang          Use Gerhard's transferable RouteHandle
-!
+!> @file
+!> @brief The FV3 atmosphere grid component nuopc cap.
+!> @author Jun Wang @date 01/2017
 
+!> @brief The FV3 atmosphere grid component nuopc cap.
+!>
+!> FV3 ATM solo model
+!>
+!> ## Module History
+!> Date | Author | Modification
+!> -----|--------|-------------
+!> 11 Oct 2016 | J. Wang | Initial code
+!> 18 Apr 2017 | J. Wang | set up fcst grid component and write grid components
+!> 24 Jul 2017 | J. Wang | initialization and time stepping changes for coupling
+!> 02 Nov 2017 | J. Wang | Use Gerhard's transferable RouteHandle
+!>
+!> @author Jun Wang @date 01/2017
 module fv3atm_cap_mod
 
   use ESMF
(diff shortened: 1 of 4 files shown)


edwardhartnett (approved):
Looks good!

Comment changes only in this PR, no code changes.

@junwang-noaa or @DusanJovic-NOAA can you help by telling us what the "???" should be?

moving_nest/fv_moving_nest_main.F90
```diff
@@ -18,1150 +22,1219 @@
...
-!! @author W. Ramstrom, AOML/HRD   05/27/2021
-!! @email William.Ramstrom@noaa.gov
-! =======================================================================!
-
+!> @brief Provides top-level interface for moving nest functionality.
+!>
+!> @author W. Ramstrom, AOML/HRD (William.Ramstrom@noaa.gov) @date  05/27/2021
 module fv_moving_nest_main_mod
 
-#include <fms_platform.h>
```
edwardhartnett, line 30: Why are there so many code changes in this file?

AlysonStahl-NOAA: I'm trying to investigate why it's showing up this way. I double checked this file in your repo and compared it to develop. It showed zero code changes, but after swapping the files for some reason it's showing up as huge chunks of changes. I followed the same procedure as the other 2 files.

AlysonStahl-NOAA: Not sure what happened, but it's fixed now.

fv3_cap.F90
```diff
@@ -56,30 +60,61 @@ module fv3atm_cap_mod
...
+  !> ???
   type(ESMF_State)                            :: fcstState
+
+  !> ???
   type(ESMF_FieldBundle), allocatable         :: fcstFB(:)
+
+  !> ???
   integer,dimension(:), allocatable           :: fcstPetList
+
+  !> ???
```
DusanJovic-NOAA, line 75: All these variables are local (private) variables and are not accessible outside the module. Are you going to add doxygen the every local variable in every module and subroutine? I thought we are adding doxygen just to subroutine arguments and public module variables.

edwardhartnett: We will not be adding doxygen for any subprogram variables. 

Good question about private module variables, I will investigate and get back to you. My hunch is that doxygen doesn't distinguish between different types of module variables.

We need to document all the module variables that doxygen will complain about if undocumented. This is so we can turn on warning-check and ensure that no undocumented code is committed to the repo.

moving_nest/fv_moving_nest_main.F90
edwardhartnett: Much better, thanks!

edwardhartnett:
@AlexanderRichert-NOAA can you get the fv3atm CI working again?

AlexanderRichert-NOAA:
@DusanJovic-NOAA what's the status of the mpi_f08 issue, did you work out a fix?

DusanJovic-NOAA:
@AlexanderRichert-NOAA  Yes, mpi_f08 will be fixed in #805, which is merged with #811 which will be merged soon. But unfortunately there was an update in upp which is causing workflow to fail now at:

sed -i 's/doc /upp_doc /' upp/docs/CMakeLists.txt

So, we'll need one more update. Do you want to fix this (remove that sed command) in your 'Add unit testing' PR?

edwardhartnett:
It's not a rush, if a fix is coming we can wait for it...

I just wanted to get the CI working again so we can start running doxygen in it.

AlexanderRichert-NOAA:
I would recommend fixing the sed failure in #811, just so we don't keep merging PRs where the CI isn't working. That said, I agree it should be as simple as removing that line since the custom target in UPP now has the name 'upp_doc' already.

DusanJovic-NOAA:
ufs-wm code managers already started or actually finished regression testing, so that change must go in as part of the next PR.

DusanJovic-NOAA:
@AlexanderRichert-NOAA @edwardhartnett I removed that sed command from fv3atm workflow file that renames  'doc' to  'upp_doc' in the UPP's CMakeLists.txt but the workflow is still failing:

https://github.com/NOAA-EMC/fv3atm/actions/runs/8833224366/job/24252100176#step:6:220

There is a mismatch in the name of doc folders in UPP version fv3atm is using. I see that is fixed 3 days ago in UPP https://github.com/NOAA-EMC/UPP/pull/932/files

@WenMeng-NOAA Looks like we will need another fv3atm PR that updates UPP submodule.

PR #848, opened 2024-06-13
Refactor NRL h2o photochemical scheme

dustinswales:
## Description
Included in this PR are changes to NRL's stratospheric h2o photochemistry scheme to make it "safe across multiple CCPP instances". These changes are identical to the reorganization that [o3 photochemistry scheme underwent in the fall of 2023.](https://github.com/ufs-community/ccpp-physics/pull/75). 

Both schemes are called from a new module, GFS_photochemistry, at the start of the time-split section of the suite. Previously, the ozone scheme was called from GFS_stateout_update, and the h2o scheme was called just after.

### Issue(s) addressed
N/A

## Testing

No changes to RT's on Hera with Intel

### DIFF
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -5334,6 +5334,12 @@
   units = mixed
   dimensions = ()
   type = ty_ozphys
+[h2ophys]
+  standard_name = dataset_for_h2o_photochemistry_physics
+  long_name = dataset for NRL h2o photochemistry physics
+  units = mixed
+  dimensions = ()
+  type = ty_h2ophys
 [h2o_phys]
   standard_name = flag_for_stratospheric_water_vapor_physics
   long_name = flag for stratospheric water vapor physics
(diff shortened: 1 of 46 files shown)


grantfirl:
@dustinswales I'm not seeing a UFS WM PR for this. Shouldn't there be one? Or were you expecting this to get combined with something since it doesn't change results?

dustinswales:
@grantfirl I will update my branches and open a PR into UWM.

ccpp/data/GFS_typedefs.meta
```diff
@@ -4821,6 +4821,11 @@
   units = count
   dimensions = ()
   type = integer
+[thpsnmp_is_init]
+  standard_name = flag_for_thompson_mp_scheme_initialization
+  long_name = flag carrying scheme initialization status
+  units = flag
```
grantfirl, line 4827: @dustinswales FYI, this variable is missing the required dimensions attribute. I'm currently running RTs for a combined PR for this and Lisa's, so you probably don't need to fix anything since it is already fixed in the combined branch. I'm only commenting in case we can't keep the combination for some reason.

grantfirl:
@jkbk2004 @dustinswales This can be closed since it is combined into https://github.com/NOAA-EMC/fv3atm/pull/865

PR #874, opened 2024-10-08
Update calculation of time-averaged radiation variables

LarissaReames:
## Description

This PR fixes incorrect diagnostic output for fhzero-averaged radiation output variables. Presently, there is a significant "sawtooth" appearance to these fields (such as dswrf_ave and ulwrf_avetoa) when we expect smoother lines with jumps only at fhzero diagnostic reset times. Here is an example of the present (OLD) result and the result from the current PR (FIX) from a 24-hour S2S regression test forecast.

iTerm2 RaUnsB rads_fhzero6_fix

### Issue(s) addressed

- fixes an problem found in investigating ufs-weather-model/[#1767](https://github.com/ufs-community/ufs-weather-model/issues/1767)

## Testing

Ran full regression test suite on Hera. See log files here: /scratch1/NAGAPE/hpc-wof1/lreames/ufs-weather-model-38a29a6/tests/logs/log_hera
All regression tests fail as sfcf*.nc files are not identical.  New baselines will be necessary.

## Dependencies

N/A. All changes are contained in fv3atm.

# Requirements before merging
- [ ] All new code in this PR is tested by at least one unit test
- [ ] All new code in this PR includes Doxygen documentation
- [ ] All new code in this PR does not add new compilation warnings (check CI output)

### DIFF
--- a/atmos_model.F90
+++ b/atmos_model.F90
@@ -975,6 +975,7 @@ subroutine update_atmos_model_state (Atmos, rc)
 !--- local variables
   integer :: i, localrc, sec_lastfhzerofh
   integer :: isec, seconds, isec_fhzero
+  integer :: datm_temp
   logical :: tmpflag_fhzero
   real(kind=GFS_kind_phys) :: time_int, time_intfull
 !
@@ -1021,11 +1022,11 @@ subroutine update_atmos_model_state (Atmos, rc)
       endif
       if (mpp_pe() == mpp_root_pe()) write(6,*) ' gfs diags time since last bucket empty: ',time_int/3600.,'hrs'
       call atmosphere_nggps_diag(Atmos%Time)
+      call get_time ( Atmos%Time_step, dtatm_temp)
       call fv3atm_diag_output(Atmos%Time, GFS_Diag, Atm_block, GFS_control%nx, GFS_control%ny, &
                             GFS_control%levs, 1, 1, 1.0_GFS_kind_phys, time_int, time_intfull, &
-                            GFS_control%fhswr, GFS_control%fhlwr)
+                            GFS_control%fhswr, GFS_control%fhlwr, datm_temp)
     endif
-
     !---  find current fhzero
     if( GFS_Control%fhzero_array(1) > 0. ) then
       fhzero_loop: do i=1,size(GFS_Control%fhzero_array)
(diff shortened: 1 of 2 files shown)


DeniseWorthen:
@LarissaReames-NOAA Will this have any impact on exported SW fields (``dnirbmi_cpl`` etc)?

LarissaReames:
> @LarissaReames-NOAA Will this have any impact on exported SW fields (`dnirbmi_cpl` etc)?

This should only apply to history file diagnostic fields with time_avg_kind = "rad_sw", "rad_lw", or "rad_swlw_min" in ccpp/driver/GFS_diagnostics.F90

LarissaReames:
@junwang-noaa Who could we have review this PR?

atmos_model.F90
```diff
@@ -975,6 +975,7 @@ subroutine update_atmos_model_state (Atmos, rc)
 !--- local variables
   integer :: i, localrc, sec_lastfhzerofh
   integer :: isec, seconds, isec_fhzero
+  integer :: datm_temp
```
junwang-noaa: maybe the previous name dtatm_temp is better? We have a data atm model called datm.

io/fv3atm_history_io.F90
```diff
@@ -445,7 +455,43 @@ subroutine history_type_output(hist, time, diag, atm_block, nx, ny, levs, ntcw,
...
+                        this_radfh = time_radlw
+                      case ('rad_swlw_min')
+                        this_radfh = min(time_radsw,time_radlw)
+                      case default
+
+                    end select
+
+                    if (time_int == dt_atmos ) then
+                      var2(i,j) = Diag(idx)%data(nb)%var2(ix)*lcnvfac
+                      hist%rad_accum(idx,ii,jj) = Diag(idx)%data(nb)%var2(ix)
```
junwang-noaa, line 472: The computation of accumulated radiation or physics stays in CCPP physics (fluxr), I suggest making the accumulation related change in ccpp, not in the IO file (fv3atm_history_io.F90). We only apply time scale or mask update in the fv3atm_history_io.F90.

Qingfu-Liu: I looked the CCPP code, and see that the fluxr is accumulated every radiation hour (3600s) (multiplied by coszdg(i) / coszen(i)) in the code:
physics/physics/Interstitials/UFS_SCM_NEPTUNE/GFS_rrtmg_post.F90
so we need modify the bucket weighting in order to calculate the dswrf_ave and ulwrf correctly

PR #885, opened 2024-10-31
Minor modification and bug fix in GFS cumulus convection schemes

rhaesung:
## Description

This PR mainly involves changes in ccpp-physics, and the code was provided by @JongilHan66.

1. Modified prognostic updraft fraction (sigmab) calculation in 'progsigma_calc.f90' which is physically more sound:
    a) moisture convergence calculation: integrate from the convection source level rather than from the cloud base
    b) 2D advection of sigmab:  sigmab advection averaged over the cloud layers rather than taking a maximum sigmab advection from k=2 to the model top
   c) To suppress unrealistically large reflectivity in the model first time step, minimum sigmab at the first time step is set to zero

2. Fix in missing vertical transport of turbulent kinetic energy (TKE) when aerosol transport is turned on (samfdeepcnv.f & samfshalcnv.f)

3. Introduce TKE at model layer interfaces (tkeh) for use in convection schemes (CCPP_typedefs.F90, CCPP_typedefs.meta, satmedmfvdifq.F, satmedmfvdifq.meta, samfdeepcnv.f, samfdeepcnv.meta, samfshalcnv.f, and samfshalcnv.meta)

4. Vertical transports of hydrometeor variables are currently not allowed in the convection schemes. But vertical transports of number concentrations of only cloud water and ice are mistakenly allowed, which is fixed in this update (CCPP_typedefs.F90)

5. All the modifications and bug fixes above had neutral impacts on the GFS forecasts

## Dependencies

- waiting on ufs-community/ccpp-physics  #https://github.com/ufs-community/ccpp-physics/pull/232

### DIFF
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1962,6 +1962,9 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: tsnowpb(:)     => null()   !< accumulated surface snowfall in bucket (m)
     real (kind=kind_phys), pointer :: rhonewsn1(:)   => null()   !< precipitation ice density outside RUC LSM (kg/m3)
 
+    !--- TKE used by convection schemes
+    real (kind=kind_phys), pointer :: tkeh(:,:)      => null()   !< vertical turbulent kinetic energy (m2/s2) at the model layer interfaces
+
     !--- MYNN variables
     real (kind=kind_phys), pointer :: edmf_a     (:,:)   => null()  !
     real (kind=kind_phys), pointer :: edmf_w     (:,:)   => null()  !
@@ -7898,6 +7901,9 @@ module GFS_typedefs
     allocate (Diag%refl_10cm(IM,Model%levs))
     allocate (Diag%max_hail_diam_sfc(IM))
 
+    !--- Vertical turbulent kinetic energy at model layer interfaces
+    allocate (Diag%tkeh(IM,Model%levs))
+
     !--- New PBL Diagnostics
     allocate (Diag%dkt(IM,Model%levs))
     allocate (Diag%dku(IM,Model%levs))
@@ -8246,6 +8252,9 @@ module GFS_typedefs
 
 !
 !-----------------------------
+
+! Vertical turbulent kinetic energy at modle layer interfaces
+    Diag%tkeh = zero
 
 ! Extra PBL diagnostics
     Diag%dkt = zero
--- a/ccpp/data/GFS_typedefs.F90
+++ b/ccpp/data/GFS_typedefs.F90
@@ -1993,9 +1993,6 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: tsnowpb(:)     => null()   !< accumulated surface snowfall in bucket (m)
     real (kind=kind_phys), pointer :: rhonewsn1(:)   => null()   !< precipitation ice density outside RUC LSM (kg/m3)
 
-    !--- TKE used by convection schemes
-    real (kind=kind_phys), pointer :: tkeh(:,:)      => null()   !< vertical turbulent kinetic energy (m2/s2) at the model layer interfaces
-
     !--- MYNN variables
     real (kind=kind_phys), pointer :: edmf_a     (:,:)   => null()  !
     real (kind=kind_phys), pointer :: edmf_w     (:,:)   => null()  !
@@ -7955,9 +7952,6 @@ module GFS_typedefs
     allocate (Diag%refl_10cm(IM,Model%levs))
     allocate (Diag%max_hail_diam_sfc(IM))
 
-    !--- Vertical turbulent kinetic energy at model layer interfaces
-    allocate (Diag%tkeh(IM,Model%levs))
-
     !--- New PBL Diagnostics
     allocate (Diag%dkt(IM,Model%levs))
     allocate (Diag%dku(IM,Model%levs))
@@ -8294,9 +8288,6 @@ module GFS_typedefs
 
 !
 !-----------------------------
-
-! Vertical turbulent kinetic energy at modle layer interfaces
-    Diag%tkeh = zero
 
 ! Extra PBL diagnostics
     Diag%dkt = zero
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -9378,7 +9378,7 @@
   long_name = vertical turbulent kinetic energy at model layer interfaces
   units = m2 s-2
   dimensions = (horizontal_loop_extent,vertical_layer_dimension)
-  type = real 
+  type = real
   kind = kind_phys
 [dkt]
   standard_name = atmosphere_heat_diffusivity
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -9440,13 +9440,6 @@
   dimensions = (horizontal_dimension)
   type = real
   kind = kind_phys
-[tkeh]
-  standard_name = vertical_turbulent_kinetic_energy_at_interface
-  long_name = vertical turbulent kinetic energy at model layer interfaces
-  units = m2 s-2
-  dimensions = (horizontal_loop_extent,vertical_layer_dimension)
-  type = real
-  kind = kind_phys
 [dkt]
   standard_name = atmosphere_heat_diffusivity
   long_name = atmospheric heat diffusivity
--- a/ccpp/data/GFS_typedefs.meta
+++ b/ccpp/data/GFS_typedefs.meta
@@ -9373,6 +9373,13 @@
   dimensions = (horizontal_dimension)
   type = real
   kind = kind_phys
+[tkeh]
+  standard_name = vertical_turbulent_kinetic_energy_at_interface
+  long_name = vertical turbulent kinetic energy at model layer interfaces
+  units = m2 s-2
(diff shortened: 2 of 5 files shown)


JongilHan66 (approved):

RuiyuSun (approved):

ccpp/data/GFS_typedefs.F90
```diff
@@ -1993,6 +1993,9 @@ module GFS_typedefs
     real (kind=kind_phys), pointer :: tsnowpb(:)     => null()   !< accumulated surface snowfall in bucket (m)
     real (kind=kind_phys), pointer :: rhonewsn1(:)   => null()   !< precipitation ice density outside RUC LSM (kg/m3)
 
+    !--- TKE used by convection schemes
+    real (kind=kind_phys), pointer :: tkeh(:,:)      => null()   !< vertical turbulent kinetic energy (m2/s2) at the model layer interfaces
```
grantfirl: @rhaesung I'm wondering why the Diag DDT was chosen to house this variable. Isn't it basically a variable that needs to get passed from PBL to deep convection and not retained from one timestep to the next? If so, it probably belongs in the CCPP_typedefs/Interstitial DDT. I doubt that this is causing the reproducibility problem, but all variables held in the Diag datatype are subject to clearing on the bucket interval (nszero). This isn't necessary because it is defined as intent(out) in satmedmfvdifq.F90 and overwritten every timestep anyway.

ccpp/data/GFS_typedefs.meta
```diff
@@ -9440,6 +9440,13 @@
   dimensions = (horizontal_dimension)
   type = real
   kind = kind_phys
+[tkeh]
+  standard_name = vertical_turbulent_kinetic_energy_at_interface
+  long_name = vertical turbulent kinetic energy at model layer interfaces
+  units = m2 s-2
+  dimensions = (horizontal_loop_extent,vertical_layer_dimension)
```
grantfirl: @rhaesung There is a mismatch between the standard name and the vertical dimension. Variables defined on the interfaces between model layers should use the standard name of `vertical_interface_dimension`.

ccpp/data/GFS_typedefs.F90
```diff
@@ -7952,6 +7955,9 @@ subroutine diag_create (Diag, Model)
     allocate (Diag%refl_10cm(IM,Model%levs))
     allocate (Diag%max_hail_diam_sfc(IM))
 
+    !--- Vertical turbulent kinetic energy at model layer interfaces
+    allocate (Diag%tkeh(IM,Model%levs))
```
grantfirl: Shouldn't this be Model%levs+1 if it is truly defined on the interfaces?

ccpp/data/GFS_typedefs.F90
rhaesung: @JongilHan66 Could you help address this?

ccpp/data/GFS_typedefs.F90
JongilHan66: @grantfirl @rhaesung   Although tkeh is tke at model interfaces, I think it can has vertical dimension of levs rather than levs+1 because it's vertical dimension is defined as "vertical_layer_dimension" consistently in all relevant meta files (i.e., GFS_typedefs.meta, satmedmfvdifq.meta, samfdeepcnv.meta, samfshalcnv.meta). Am I wrong? If so, Rhaesung, please change "vertical_layer_dimension" to "vertical_interface_dimension" for tkeh in all the meta files and also change "Diag%tkeh(IM,Model%levs)" to "Diag%tkeh(IM,Model%levs+1)" in GFS_typedefs.F90.  Once the prognostic tke is updated every time step, then tkeh is computed as tkeh=0.5*(tke(k)+tke(k+1)) up to k=levs-1 every time step in satmedmfvdifq.F for use in samfdeepcnv.f and samfshalcnv.f.

grantfirl: @JongilHan66 You're right that the consistency of naming across all files is correct, but I think that it is at the very least confusing to have a variable intended for use on interfaces to be given the vertical dimension corresponding to the model layer centers. If you look at other variables defined on interfaces, they're given the vertical dimension of `vertical_interface_dimension`. I'm just looking at the changes in this PR to try to help debug the unreproducible regression tests, and I think that uninitialized sections of arrays have the tendency to create unreproducible code. When `tkeh` is calculated in satmedmfvdifq.F, I think that it would make sense to make sure that the entire array is given a value, not just up to k=levs-1, just in case that the array is ever accessed outside levs-1 in subsequent schemes. Even setting the entire tkeh array to 0 before doing the loop from k=1 to levs-1 would fix the potential problem.

JongilHan66: @grantfirl @rhaesung  Grant, tkeh is initialized to be zero in GFS_typedefs.F90 as "Diag%tkeh = zero". Do we still need to initialize in satmedmfvdifq.F?

JongilHan66: @rhaesung Rhaesung, could you also set tkeh = 0. before doing the loop from k=1 to levs-1 in satmedmfvdifq.F and test it if it can fix the problem?

grantfirl: Yes, because it is intent(out) in satmedmfvdifq. Any time a variable is given intent(out) in a scheme, it is (potentially) given garbage values. Plus, variables in the Diag DDT are cleared out and reinitialized every nszero timesteps, which is unnecessary. Another "fix" would be to just make tkeh intent(inout). That way, the initialization within diag_phys_zero would at least be kept, although I'd argue it's still being held in the wrong DDT for this application. It should be in CCPP_typedefs.F90/Interstitial DDT that get reinitialized every physics timestep.

JongilHan66: @grantfirl @rhaesung Thanks, Grant for the explanation!!
Rhaesung, before you try to change from "vertical_layer_dimension" to "vertical_interface_dimension",
please do these tests in order as:
1) make tkeh intent(inout) in all the 4 meta files
2) set tkeh=0 before doing the loop from k=1 to levs-1 in satmedmfvdifq.F

rhaesung: @JongilHan66 Sure, I'll do that and report back to you.

rhaesung: @JongilHan66 @grantfirl Unfortunately, neither of them is working to resolve the issue. I’ll determine if the issue is actually related to `tkeh` and try to follow Grant's suggestion.

rhaesung: @JongilHan66 @grantfirl Okay, I was able to compile and test the run with tkeh removed, but the baseline reproduction issue still persists. So, it seems the issue isn't related to tkeh. @grantfirl Do you see any other issues aside from tkeh?

ccpp/data/GFS_typedefs.F90
rhaesung: @grantfirl @JongilHan66 Thanks for bringing this up! After further investigation, it turned out that this variable is indeed more appropriate for the CCPP_typedefs/Interstitial DDT to pass some of the failed tests (e.g., hafs_regional_storm_following_1nest_atm_ocn,...). I’ve made this change accordingly.

ccpp/data/GFS_typedefs.meta
rhaesung: @grantfirl Thanks, done!

jkbk2004:
@rhaesung can you point to https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/tree/remap_io_emc? https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/374 has no impact on weather model baselines. So it's good to bring it in.

rhaesung:
> @rhaesung can you point to https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/tree/remap_io_emc? [NOAA-GFDL/GFDL_atmos_cubed_sphere#374](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/374) has no impact on weather model baselines. So it's good to bring it in.

@jkbk2004 Done!

grantfirl:
@rhaesung https://github.com/ufs-community/ccpp-physics/pull/232 is merged. Please update the submodule pointer for ccpp-physics (https://github.com/ufs-community/ccpp-physics/commit/435e80abcd386d873bd4d0a03fbf7deffeaf9528) and revert .gitmodules in preparation for merge.

rhaesung:
> @rhaesung [ufs-community/ccpp-physics#232](https://github.com/ufs-community/ccpp-physics/pull/232) is merged. Please update the submodule pointer for ccpp-physics ([ufs-community/ccpp-physics@435e80a](https://github.com/ufs-community/ccpp-physics/commit/435e80abcd386d873bd4d0a03fbf7deffeaf9528)) and revert .gitmodules in preparation for merge.

@grantfirl Thanks, done!

jkbk2004:
@rhaesung can you update cubed sphere hash: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/commit/127a341d126a6225a52eed61c4e866e1af82005f ?

rhaesung:
> @rhaesung can you update cubed sphere hash: [NOAA-GFDL/GFDL_atmos_cubed_sphere@127a341](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/commit/127a341d126a6225a52eed61c4e866e1af82005f) ?

@jkbk2004 Sure, done!

jkbk2004 (approved):

DusanJovic-NOAA (approved):

PR #909, opened 2025-01-15
Enable customizable idealized, doubly-periodic tests

### LINKED ISSUE #2558: Create a user-friendly idealized testing capability
## Description
When testing new (or old) code, or trying to debug odd model behavior, especially in physics development, it's helpful to have a cheap, simplified idealized capability. While there are some hard-coded test cases in atmos_cubed_sphere, many are disused or completely non-functional in the UFS framework. 

## Solution
Implement a new, run-time customizable idealized test on a doubly-periodic domain that can be run with CCPP physics and has the option of reading in external soundings (akin to WRF's ideal.exe) to initialized model state. This capability has been in use at NSSL for a few years, so it's just a matter of fixing up the code and contributing it back to the community. 

## Alternatives
N/A

## Related to
N/A

cc @MicroTed as he's been instrumental in working on this capability with me

LarissaReames:
## Description

This PR adds a doubly-periodic idealized test case to test_cases.F90 that is customizable by the user at run time via input.nml and that can be run within the UFS framework and is fully compatible with CCPP physics. 

An important modification was made to make this case compatible with CCPP. To enable the user to choose the canned  "SupeCell" sounding from GFDL's original supercell test, we copied over the routines from GFDL MP that were previously preventing this code from compiling with CCPP/"GFS physics". 

We have added documentation to help the user, including some code to generate an idealized surface file that the user can customized to their liking. The PR at the UWM level will also have a corresponding regression test. It won't be baseline tested, but it will provide an easy route for user to generate a run directory. 

Many thanks go to @MicroTed for all of his help and contributions to this effort as well.  He's agreed to provide a test run directory repo that can be cloned by those who'd like to test the code inside a pre-generated run directory.  You can find that here: https://github.com/MicroTed/ufs_ideal_run.git

### Issue(s) addressed

- fixes ufs-community/ufs-weather-model/issues/2558

## Testing

Tested on Hera (intel and gnu, release and debug builds for both) and on Jet. Tests have been run to check that all the various types of sounding options can be used successfully. 

The full suite of regression tests will be run at the weather model level. A regression test will be added in the WM. 

No existing baseline changes are expected. 

## Dependencies

- waiting on NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/370

# Requirements before merging
- [ ] All new code in this PR is tested by at least one unit test
- [ ] All new code in this PR includes Doxygen documentation
- [ ] All new code in this PR does not add new compilation warnings (check CI output)


LarissaReames:
This PR doesn't add any new subroutines or inputs for routines contained in fv3atm, but I did add an if statement branch specifically for idealized test cases (grid_type=4)  in module_fcst_grid_comp.F90 that would probably need a unit test. I don't think this file is covered by any unit tests as of yet, @edwardhartnett how do you want to proceed here? Do we need a unit test that test all branches of that if block?

edwardhartnett:
Don't worry about the unit testing yet.

module_fcst_grid_comp.F90
```diff
@@ -61,7 +61,9 @@ module module_fcst_grid_comp
                                     fv_srf_wnd_restart_bundle_setup, &
                                     fv_tracer_restart_bundle_setup
 
-  use fms2_io_mod,        only: FmsNetcdfFile_t, open_file, close_file, variable_exists, read_data
+  use fms2_io_mod,        only: FmsNetcdfFile_t, open_file, close_file, variable_exists, & 
+                                read_data
+  use fms_io_mod,         only: field_exist, read_data_old=>read_data
```
DusanJovic-NOAA: Why is `read_data` subroutine from an old fms_io_mod needed? Can't `read_data` from  fms2_io_mod be used to read gridspec file used by these idealized tests?

module_fcst_grid_comp.F90
```diff
@@ -237,41 +262,43 @@ subroutine SetServicesNest(nest, rc)
...
-    array = ESMF_ArrayCreate(distgrid, farray=Atmos%lat_bnd, indexflag=ESMF_INDEX_DELOCAL, rc=rc)
-    if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
-
-    call ESMF_GridSetCoord(grid, coordDim=2, staggerLoc=ESMF_STAGGERLOC_CORNER, array=array, rc=rc)
-    if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
-
-    !TODO: Consider aligning mask treatment with coordinates... especially if it requires updates for moving
-    call addLsmask2grid(grid, rc=rc)
-    if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
+    if (.not. (Atmos%grid_type==4)) then
```
DusanJovic-NOAA: I would write this if tests as:

`if (Atmos%grid_type /= 4) then`

Also, why grid with grid_type == 4 does not need grid coordinates, as other grid types? Is grid_type == 4 special in that regard?

module_fcst_grid_comp.F90
LarissaReames: Good catch. This must have been a quick patch in testing. I've replaced its use with the standard read_data from fms2_io after testing confirmed it compiles and runs as normal.

module_fcst_grid_comp.F90
LarissaReames: So originally this was bypassed because the ESMF coordinates have no bearing when the write component grid is never used, and it can't be used with these doubly periodic domains as they can have unrealistic (e.g., uniform) lat/lon configurations. However, I've pushed a fix that creates a single tile non-periodic ESMF grid for grid_type==4, and this now fills the ESMF coordinates correctly even if they're never used here.

module_fcst_grid_comp.F90
```diff
@@ -201,7 +203,37 @@ subroutine SetServicesNest(nest, rc)
...
-    if (trim(name)=="global") then
+    if (Atmos%grid_type==4) then
+       gridfile = "grid_spec.nc" ! default
+
+       if (open_file(fileobj, "INPUT/grid_spec.nc", "read")) then
+         if (variable_exists(fileobj, "atm_mosaic_file")) then
+           call read_data(fileobj, "atm_mosaic_file", gridfile)
+         endif
+         call close_file(fileobj)
+       endif
```
DusanJovic-NOAA: Looks to me `gridfile` is not used anymore, other than in the error message below. If it's not, it should be removed.

module_fcst_grid_comp.F90
```diff
@@ -201,7 +203,37 @@ subroutine SetServicesNest(nest, rc)
...
+                                      coordSys=ESMF_COORDSYS_SPH_RAD, &
+                                      coordTypeKind=grid_typekind, &
+                                      decompflag=(/ESMF_DECOMP_SYMMEDGEMAX,ESMF_DECOMP_SYMMEDGEMAX/), &
+                                      name="fcst_grid", &
+                                      indexflag=ESMF_INDEX_DELOCAL, &  
+                                      rc=rc) 
+      if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU,line=__LINE__, file=__FILE__)) return
+      call mpp_error(NOTE, 'after create fcst grid for doubly periodic with INPUT/'//trim(gridfile))
+
+      call addLsmask2grid(grid, rc=rc)
```
DusanJovic-NOAA: Is this call needed here? There is already the same call on line 306 below.

module_fcst_grid_comp.F90
```diff
@@ -201,7 +203,37 @@ subroutine SetServicesNest(nest, rc)
...
+      grid = ESMF_GridCreateNoPeriDim( regDecomp=(/layout(1),layout(2)/), &
+                                      minIndex=(/1,1/), &
+                                      maxIndex=(/nx,ny/), &
+                                      gridAlign=(/-1,-1/), &
+                                      coordSys=ESMF_COORDSYS_SPH_RAD, &
+                                      coordTypeKind=grid_typekind, &
+                                      decompflag=(/ESMF_DECOMP_SYMMEDGEMAX,ESMF_DECOMP_SYMMEDGEMAX/), &
+                                      name="fcst_grid", &
+                                      indexflag=ESMF_INDEX_DELOCAL, &  
+                                      rc=rc)
```
DusanJovic-NOAA: This block looks to be identical to the code in `else` block below. If it is indeed identical, these two cases should be combined into one.

LarissaReames: Logic has been combined for the two cases.

module_fcst_grid_comp.F90
LarissaReames: Call removed.

module_fcst_grid_comp.F90
LarissaReames: This block has been removed.

MicroTed:
New PR at https://github.com/NOAA-EMC/fv3atm/pull/953 to replace this one

DusanJovic-NOAA:
PR #953 has been merged. Closing.

PR #916, opened 2025-01-27
CCPP physics: wet() bug fix and workaround for bad tsfco

SamuelTrahanNOAA:
## Description

Fixes this problem:

- https://github.com/ufs-community/ufs-weather-model/issues/2547

It was caused by missing values (-1e+20) for sea surface temperature on `wet` points in input data. (`Wet` points are points where the CCPP `wet` array is true, thus indicating the surface is at least partly water.) There's a bug fix and a workaround:

1. Bug fix: In certain situations, the `wet` array can be mistakenly true when it should be false due to the location being 100% ice. An "else" statement corrects this.
2. Workaround: If the input sea surface temperatures (tsfco) are invalid on a `wet` point, the lowest model level temperature is used instead.

There is no other temperature available on those points that can reasonably be used as a sea surface temperature. The lowest model level temperature may not be a perfect analogue for the sea surface temperature, but it'll be more realistic than -1E+20 K.

### Issue(s) addressed

- https://github.com/ufs-community/ufs-weather-model/issues/2547

- Confirmed that infinities in tendencies go away.
- Ran hera regression tests to ensure no tests change.

Hera.

A lower-resolution version of this configuration has a regression test.

This specific input that causes a problem is too high a resolution to be a regression test. Also, this problem likely originates from trouble with the initial conditions. We don't need to test the model with potentially problematic input.

Hera.

No.

## Dependencies

- waiting on https://github.com/ufs-community/ccpp-physics/pull/247

# Requirements before merging
- [x] All new code in this PR is tested by at least one unit test
- [x] All new code in this PR includes Doxygen documentation (no documentation changes needed)
- [x] All new code in this PR does not add new compilation warnings (check CI output)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,8 +8,8 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/ufs-community/ccpp-physics
-  branch = ufs/dev
+  url = https://github.com/SamuelTrahanNOAA/ccpp-physics
+  branch = bugfix/good-wet-tsfco
 [submodule "upp"]
   path = upp
   url = https://github.com/NOAA-EMC/UPP


grantfirl:
@DusanJovic-NOAA This was combined into https://github.com/NOAA-EMC/fv3atm/pull/896 so it can be closed when you get a chance.

PR #918, opened 2025-01-28
Fixing some doxygen documentation - first pass

edwardhartnett:
## Description

Fixing some doxygen documentation

### Issue(s) addressed

- Part of #760

## Testing

## Dependencies

### DIFF
--- a/cpl/module_block_data.F90
+++ b/cpl/module_block_data.F90
@@ -1,3 +1,6 @@
+!> @file
+!> @brief ???
+!> @author ???
 module module_block_data
 
   ! Copies block data containing real*4, real*8, or integer into
@@ -71,8 +74,17 @@ module module_block_data
 
 contains
 
-  ! -- copy: 1D to 2D
-
+  !> copy: 1D to 2D ???
+  !> 
+  !> @param[in] destin_ptr ???
+  !> @param[in] source_ptr ???
+  !> @param[in] block ???
+  !> @param[in] block_index ???
+  !> @param[in] scale_factor ???
+  !> @param[in] offset ???
+  !> @param[out] rc ???
+  !>
+  !> @author
   subroutine block_copy_1d_i4_to_2d_r8(destin_ptr, source_ptr, block, block_index, scale_factor, offset, rc)
 
     ! -- arguments
@@ -110,6 +122,17 @@ contains
 
   end subroutine block_copy_1d_i4_to_2d_r8
 
+  !> ???
+  !> 
+  !> @param[in] destin_ptr ???
+  !> @param[in] source_ptr ???
+  !> @param[in] block ???
+  !> @param[in] block_index ???
+  !> @param[in] scale_factor ???
+  !> @param[in] offset ???
+  !> @param[out] rc ???
+  !>
+  !> @author
   subroutine block_copy_1d_r8_to_2d_r8(destin_ptr, source_ptr, block, block_index, scale_factor, offset, rc)
 
     ! -- arguments
@@ -147,6 +170,18 @@ contains
 
   end subroutine block_copy_1d_r8_to_2d_r8
 
+  !> ???
+  !> 
+  !> @param[in] destin_ptr ???
+  !> @param[in] source_ptr ???
+  !> @param[in] block ???
+  !> @param[in] block_index ???
+  !> @param[in] scale_factor ???
+  !> @param[in] special_value ???
+  !> @param[in] offset ???
+  !> @param[out] rc ???
+  !>
+  !> @author
   subroutine block_copy_spval_1d_r8_to_2d_r8(destin_ptr, source_ptr, block, block_index, scale_factor, special_value, offset, rc)
 
     ! -- arguments
@@ -186,8 +221,18 @@ contains
 
   end subroutine block_copy_spval_1d_r8_to_2d_r8
 
-  ! -- copy: 1D slice to 2D
-
+  !> copy: 1D slice to 2D ???
+  !> 
+  !> @param[in] destin_ptr ???
+  !> @param[in] source_ptr ???
+  !> @param[in] slice ???
+  !> @param[in] block ???
+  !> @param[in] block_index ???
+  !> @param[in] scale_factor ???
+  !> @param[in] offset ???
+  !> @param[out] rc ???
+  !>
+  !> @author
   subroutine block_copy_1dslice_r8_to_2d_r8(destin_ptr, source_ptr, slice, block, block_index, scale_factor, offset, rc)
 
     ! -- arguments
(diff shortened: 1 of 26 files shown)


cpl/module_block_data.F90
```diff
@@ -1,3 +1,6 @@
+!> @file
+!> @brief ???
```
edwardhartnett, line 2: @rmontuoro can you please fill in what all the "???" documentation should be in this file? According to github, you originally added this file.

cpl/module_cap_cpl.F90
```diff
@@ -13,9 +17,17 @@ module module_cap_cpl
   public diagnose_cplFields
   contains
 
-  !-----------------------------------------------------------------------------
-  !-----------------------------------------------------------------------------
-
+    !> ???
+    !> 
+    !> @param[in] gcomp ???
```
edwardhartnett, line 22: @junwang-noaa can you please tell us what the ??? should be in this documentation?

io/fv3atm_sfc_io.F90
```diff
@@ -27,27 +37,24 @@ module fv3atm_sfc_io
...
-    integer, public :: nvar3 = 0
-    integer, public :: nvar2r = 0
-    integer, public :: nvar2mp = 0
-    integer, public :: nvar3mp = 0
-    integer, public :: nvar2l = 0
-    integer, public :: nvar2m = 0
-    integer, public :: nvar_before_lake = 0
-
-    ! The lsoil flag is only meaningful when reading:;
+    integer, public :: nvar2o = 0 !< ???
```
edwardhartnett: @SamuelTrahanNOAA can you please tell us what the ??? in this documentation should be?

moving_nest/bounding_box.F90
```diff
@@ -64,7 +70,12 @@ subroutine fill_bbox_r4_2d(out_bbox, in_grid)
     out_bbox%je = ubound(in_grid, 2)
   end subroutine fill_bbox_r4_2d
 
-
+  !> ???
```
edwardhartnett, line 73: @wramstrom can you please tell us what should replace the ??? in this documentation?

stochastic_physics/stochastic_physics_wrapper.F90
```diff
@@ -1,62 +1,83 @@
...
-  real(kind=kind_phys), dimension(:,:,:), allocatable, save :: skebv_wts
-  real(kind=kind_phys), dimension(:,:,:), allocatable, save :: sfc_wts
-  real(kind=kind_phys), dimension(:,:,:,:), allocatable, save :: spp_wts
-
-  logical, save :: is_initialized = .false.
-  integer, save :: lsoil = -999
-  real(kind=kind_phys), dimension(:,:,:), allocatable, save :: smc
-  real(kind=kind_phys), dimension(:,:,:), allocatable, save :: stc
-  real(kind=kind_phys), dimension(:,:,:), allocatable, save :: slc
+  real(kind=kind_phys), dimension(:,:),   allocatable, save :: xlat !< ???
```
edwardhartnett: @climbfuji can you tell us what should replace the ??? in this documentation?

climbfuji: I would check GFS_typedefs.F90/GFS_typedefs.meta for the variable with the same name

edwardhartnett: This is your code right?

climbfuji: I wrote some of this many years ago, then @pjpegion and @lisa-bengtsson updated some of it, and recently I had to modify those fields when moving from blocked data structures to contiguous arrays. The code further down in the file unambiguously relates these local variables with variables in `GFS_typedefs.F90`, for example:
```
call transfer_field_to_stochastics(GFS_Control%blksz, GFS_Grid%xlat, xlat)
```
In `GFS_typedefs.F90`, https://github.com/NOAA-EMC/fv3atm/blob/2c7357b2613ca5be11a298f26a11cfcb0b05dd00/ccpp/data/GFS_typedefs.F90#L1694-L1700:

Thus, `xlat` here should be the same, I suppose.
```
!< grid latitude in radians, default to pi/2 ->
!! -pi/2 range, otherwise adj in subr called
```

DusanJovic-NOAA: I think all module variables in this file can be (and probably should be) declared as private. Doxygen should not complain about undocumented private (and local) variables.

edwardhartnett: Private and public module vars need to be documented.

edwardhartnett: I'm sure we all agree that this is important code and should be documented. In addition, NCO has documentation requirements which are not being met by this code.

@climbfuji since you are the original author, could you do a PR against the ejh_doxygen branch with the missing documentation from the module variables and all of the subroutines?

Presumably the meaning of this code is obvious to you, but that needs to be documented. We are intending to write unit tests for the code in fv3atm, and that will be mostly impossible without documentation of the code. Since you have an understanding of what this code is, that shouldn't take more than a few minutes of your time...

DusanJovic-NOAA: Can you provide a link to NCO's documentation requirements document where the requirement that all private and local variables are documented using doxygen is specified?

climbfuji: > I'm sure we all agree that this is important code and should be documented. In addition, NCO has documentation requirements which are not being met by this code.
> 
> @climbfuji since you are the original author, could you do a PR against the ejh_doxygen branch with the missing documentation from the module variables and all of the subroutines?
> 
> Presumably the meaning of this code is obvious to you, but that needs to be documented. We are intending to write unit tests for the code in fv3atm, and that will be mostly impossible without documentation of the code. Since you have an understanding of what this code is, that shouldn't take more than a few minutes of your time...

I won't have time for this for several weeks if not months, sorry. It's all in GFS_typedefs and can be taken from there if someone wants to do this in the meantime.

JacobCarley-NOAA:
Hi All - Documentation is a necessity and we need to work together and do the best we can to get this in shape. It is a part of the implementation standards, though not spelled out very clearly [see here](https://www.nco.ncep.noaa.gov/idsb/implementation_standards/ImplementationStandards.v11.0.0.pdf). Even if it were not in the standard, I think we all know that programming best-practices include documentation.

Going forward, documentation will become a necessary component of new PRs to this repository.

edwardhartnett:
@climbfuji please review PR here: https://github.com/edwardhartnett/fv3atm/pull/31

io/fv3atm_sfc_io.F90
edwardhartnett: Thanks @SamuelTrahanNOAA  for helping Larissa with this!

DusanJovic-NOAA:
This PR is replaced by #984

PR #920, opened 2025-02-02
Change BUILD_TESTING to BUILD_CI_TESTING in CMake, to allow ufs-weather-model to use BUILD_TESTING

### LINKED ISSUE #821: Adding -DBUILD_TESTING=ON or adding "include(CTest)" to CMakeLists.txt in ufs-weather-model breaks the build
I would like to add a couple of ctests at the atmos_cubed_sphere level that would feed up/be recognized at the top level of the ufs-weather-model build. However, when I add "include(CTest)" to the CMakeLists.txt file in ufs-weather-model, this sets BUILD_TESTING=ON and FV3 tries to build the ci tests and that fails. 

Can we either fix the ci section so that it builds with a normal, out-of-the-box build, or change the "BUILD_TESTING" variable to something like "BUILD_CI_TESTING"?

edwardhartnett:
## Description

BUILD_TESTING was used by @AlexanderRichert-NOAA to get the fv3atm special build working. It's necessary because fv3atm contains code that links to subcomponent code, and that doesn't build on its own.

But BUILD_TESTING is a standard CMake option that has other uses. So instead, we should use BUILD_CI_TESTING for our own home-rolled hackery to build fv3atm in the CI system.

### Issue(s) addressed

Fixes #821

## Testing

In CI

## Dependencies

None.

### DIFF
--- a/.github/workflows/GCC.yml
+++ b/.github/workflows/GCC.yml
@@ -120,7 +120,7 @@ jobs:
         export CC=mpicc
         export CXX=mpicxx
         export FC=mpif90
-        cmake ${GITHUB_WORKSPACE}/fv3atm -DBUILD_TESTING=ON ${{ matrix.cmake_opts }} ${{ env.gcov_cmake }}
+        cmake ${GITHUB_WORKSPACE}/fv3atm -DBUILD_CI_TESTING=ON ${{ matrix.cmake_opts }} ${{ env.gcov_cmake }}
         make -j2
 
     - name: run-tests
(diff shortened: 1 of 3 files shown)


DusanJovic-NOAA (approved):

AlexanderRichert-NOAA (approved):

climbfuji:
It looks like this will close https://github.com/ufs-community/ufs-weather-model/issues/2648 - thanks @DusanJovic-NOAA for the pointer.

jkbk2004:
merged with #931

PR #877, opened 2024-10-11
[production/RRFS.v1] fix for inline post when using RUC LSM

JiliDong-NOAA:
## Description

@ericaligo-NOAA found an issue where inline post does not correctly output soil temperature and moisture in grib2 when using RUC LSM. We tracked this to post_fv3 where iSF_SURFACE_PHYSICS is not updated for RUC LSM. This parameter is used in UPP SURFACE.f for soil variables output. The initial iSF_SURFACE_PHYSICS = 2 in post_fv3 and should be updated to 3 when RUC LSM is used. In this PR we set iSF_SURFACE_PHYSICS =3 when nsoil=9 and successfully output the 9 level soil variables.   

A PR to ufs-weather-model will be submitted later.

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
- fixes #<issue_number>
- fixes noaa-emc/fv3atm/issues/<issue_number>

## Testing

How were these changes tested?  
What compilers / HPCs was it tested with?  
Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)  
Have the ufs-weather-model regression test been run? On what platform?  
- Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.
- Please commit the regression test log files in your ufs-weather-model branch

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)

Do PRs in upstream repositories need to be merged first?
If so add the "waiting for other repos" label and list the upstream PRs
- waiting on noaa-emc/nems/pull/<pr_number>
- waiting on noaa-emc/fv3atm/pull/<pr_number>

# Requirements before merging
- [ ] All new code in this PR is tested by at least one unit test
- [ ] All new code in this PR includes Doxygen documentation
- [ ] All new code in this PR does not add new compilation warnings (check CI output)

PR #869, opened 2024-09-17
[production/RRFS.v1] saSAS sigmab initialization changes to reduce large initial reflectivity in RRFS and inline post bug fix for RUC LSM

JiliDong-NOAA:
## Description

1. This PR follows Jongil's suggestion and aims to reduce the large convective reflectivity caused by saSAS adjustment in the first timestep during a warm start. The issue is likely related to the inconsistency when DA updates the moisture at t but not the moisture from the previous timestep (t-36s). The moisture from the previous timestep is needed for initializing sigmab (updraft area fraction) when calculating qadv (q advection or tendency term).

The PR forces qadv to zero in the first timestep when a namelist parameter sigmab_coldstart is set to .true. It also reduces the lower limit of sigmab from 0.01 to 0.0 in the first timestep.

dependent on ccpp-physics PR:

https://github.com/ufs-community/ccpp-physics/pull/225

2. Two bug fixes for LSM soil output with inline post are added in this PR:
2.1 https://github.com/NOAA-EMC/fv3atm/pull/877
2.2 @ericaligo-NOAA identified and fixed another bug with inline post when "lsoil" is different from "lsoil_lsm" which can happen when cold starting from Noah LSM input but running forecast with RUC LSM. In the above example, lsoil=4 and lsoil_lsm=9. There is no soil variables from inline post. The fix will set the correct "nsoil" based on lsoil_lsm and recover the LSM soil output.

### Issue(s) addressed

Link the issues to be closed with this PR, whether in this repository, or in another repository.
(Remember, issues should always be created before starting work on a PR branch!)
- fixes #<issue_number>
- fixes noaa-emc/fv3atm/issues/<issue_number>

## Testing

How were these changes tested?  
What compilers / HPCs was it tested with?  
Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)  
Have the ufs-weather-model regression test been run? On what platform?  
- Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.
- Please commit the regression test log files in your ufs-weather-model branch

## Dependencies

If testing this branch requires non-default branches in other repositories, list them.
Those branches should have matching names (ideally)

Do PRs in upstream repositories need to be merged first?
If so add the "waiting for other repos" label and list the upstream PRs
- waiting on noaa-emc/nems/pull/<pr_number>
- waiting on noaa-emc/fv3atm/pull/<pr_number>

# Requirements before merging
- [ ] All new code in this PR is tested by at least one unit test
[...]

PR #926, opened 2025-02-07
Tenth reconciliation PR from production/RRFS.v1

grantfirl:
## Description

This is @JiliDong-NOAA 's work and is identical to #869

This PR follows Jongil's suggestion and aims to reduce the large convective reflectivity caused by saSAS adjustment in the first timestep during a warm start. The issue is likely related to the inconsistency when DA updates the moisture at t but not the moisture from the previous timestep (t-36s). The moisture from the previous timestep is needed for initializing sigmab (updraft area fraction) when calculating qadv (q advection or tendency term).
The PR forces qadv to zero in the first timestep when a namelist parameter sigmab_coldstart is set to .true. It also reduces the lower limit of sigmab from 0.01 to 0.0 in the first timestep.

Two bug fixes for LSM soil output with inline post are added in this PR:
2.1 https://github.com/NOAA-EMC/fv3atm/pull/877
2.2 @ericaligo-NOAA identified and fixed another bug with inline post when "lsoil" is different from "lsoil_lsm" which can happen when cold starting from Noah LSM input but running forecast with RUC LSM. In the above example, lsoil=4 and lsoil_lsm=9. There is no soil variables from inline post. The fix will set the correct "nsoil" based on lsoil_lsm and recover the LSM soil output. 

### Issue(s) addressed

None

## Testing

UFS RTs on Hera

Changes baselines for tests using saSAS deep convection

## Dependencies

https://github.com/ufs-community/ccpp-physics/pull/252

### DIFF
--- a/io/fv3atm_history_io.F90
+++ b/io/fv3atm_history_io.F90
@@ -187,8 +187,7 @@ CONTAINS
     hist%fhzero = Model%fhzero
     !   hist%ncld   = Model%ncld
     hist%ncld   = Model%imp_physics
-    hist%nsoil  = Model%lsoil
-    hist%nsoil_lsm  = Model%lsoil_lsm
+    hist%nsoil  = Model%lsoil_lsm
     hist%dtp    = Model%dtp
     hist%imp_physics  = Model%imp_physics
     hist%landsfcmdl  = Model%lsm
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,10 +8,8 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  #url = https://github.com/ufs-community/ccpp-physics
-  #branch = ufs/dev
-  url = https://github.com/grantfirl/ccpp-physics
-  branch = rrfsv1-to-ufs/dev10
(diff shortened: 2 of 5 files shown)


io/fv3atm_history_io.F90
```diff
@@ -187,8 +187,7 @@ subroutine history_type_register(hist, Diag, Time, Atm_block, Model, xlon, xlat,
     hist%fhzero = Model%fhzero
     !   hist%ncld   = Model%ncld
     hist%ncld   = Model%imp_physics
-    hist%nsoil  = Model%lsoil
-    hist%nsoil_lsm  = Model%lsoil_lsm
+    hist%nsoil  = Model%lsoil_lsm
```
grantfirl, line 190: @JiliDong-NOAA Please take a look here. I noticed that there was a difference between what exists in the develop branch and your PR into production/RRFS.v1 here: https://github.com/NOAA-EMC/fv3atm/pull/869. I'm not sure that I handled this conflict correctly.

jkbk2004 (approved):

BrianCurtis-NOAA (approved):

PR #929, opened 2025-02-10
Use ESMF Redist instead of Remap for history bundles with 'cubed_sphere_grid' output grid

### LINKED ISSUE #2589: Zeros along edges of regression test output
## Description
I have been working on developing a regional Arctic domain based off existing HAFS regression tests in the UFS weather model. I noticed that when using `cubed_sphere_grid` output, the `atmf00*.nc` files contain anomalous zeros along the edges of the domain. Looking back through the regression tests, I found that these zeros are also present in the `hafs_regional_atm_thompson_gfdlsf` regression test.

### To Reproduce:
The error can be found in the baseline dataset for `hafs_regional_atm_thompson_gfdlsf` regression test. I have been working on Hera and found the zeros were still present as recently as the 2025-01-29 baseline files (`/scratch2/NAGAPE/epic/UFS-WM_RT/NEMSfv3gfs/develop-20250129/hafs_regional_atm_thompson_gfdlsf_intel/`)

## Output
The anomolous zeros are all along the edges of the domain. For example, below is quick output showing where the `tmp` variable is equal to 0 (purple) and not 0 (yellow). This is taken from the `hafs_regional_atm_thompson_gfdlsf` regression test data. The output is plotting the `time=0` and `pfull=0` layer, but the distribution of 0s appears to be the same for every layer and for every variable in the dataset.

Image

DusanJovic-NOAA:
## Description

Fix a bug in the write grid component where regional cubed sphere output have undefined (zero) values on some points on the edge (outermost rows and columns). This happens only when model is compiled with -D32BIT=ON option due to round off differences in lat/lon grid coordinate values between forecast and write grid component. When model is compiled without -D32BIT=ON, both forecast and write grid components use 64bit values for grid coordinates and in that case during the remapping all grid points are defined (i.e. 'inside the model domain'). Solution is to use ESMF Redist instead of Remap for all cubed sphere grid outputs.

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/2589

## Testing

How were these changes tested?  Yes. UFS regression test, and Arctic domain test as reported in https://github.com/ufs-community/ufs-weather-model/issues/2589
What compilers / HPCs was it tested with?  Intel and GNU on Hercules.
Have the ufs-weather-model regression test been run? Yes. On what platform?  Hercules
- Will the code updates change regression test baseline? Yes. All regression test that output cubed sphere history files.

## Dependencies

No.

### DIFF
--- a/fv3_cap.F90
+++ b/fv3_cap.F90
@@ -788,8 +788,7 @@ module fv3atm_cap_mod
               endif
             endif
 
-            call ESMF_LogWrite('bf FieldBundleRegridStore', ESMF_LOGMSG_INFO, rc=rc)
-            write(msgString,"(A,I2.2,',',I2.2,A)") "calling into wrtFB(",j,i, ") FieldBundleRegridStore()...."
+            write(msgString,"(A,I2.2,',',I2.2,A)") "RH creation for wrtFB(",j,i, ") ...."//trim(fcstItemNameList(j))
             call ESMF_LogWrite(msgString, ESMF_LOGMSG_INFO, rc=rc)
 
             if (i==1) then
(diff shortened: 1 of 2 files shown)


jkbk2004 (approved):

jkbk2004:
Tests are done at https://github.com/ufs-community/ufs-weather-model/pull/2595. We can merge this pr.

FernandoAndrade-NOAA (approved):

PR #949, opened 2025-04-10
Add qv photochemistry diagnostics

### LINKED ISSUE #255: Add diagnostics for water vapor photochemistry parametrization
## Description

The documentation implies I can request `dtend_qv_photochem` and `dtend_qv_prodloss` from the water vapor photochemistry scheme or its wrapper, as I can for the ozone photochemistry scheme.
CCPP currently quietly skips those, leaving me to rely on `dtend_qv_phys`, which probably includes photochemistry, and is probably just photochemistry near the stratopause.

However, forecasts produce excess water vapor near and above the stratopause, and the water vapor field varies enough I'd like to be sure, and to investigate which of the terms of the parametrization produces the increase and variation.

## Solution

Wrap wherever `physics/photochem/module_h2ophys.F90` with the same bits around the call locations of `physics/photochem/module_ozphys.F90` that let it calculate `dtend_o3_photochem`, then add the array allocation and optional argument to `module_h2ophys%run` to allow it to calculate and return `dtend_qv_prodloss`.  I can't tell where those calls and allocations happen well enough to add them myself.

## Alternatives (optional)

- Run ccpp-scm for an hour at a time without advection somewhere around 75°S to find photochemical tendencies
- Run the photochemical scheme myself outside of CCPP to get the tendencies
- Point CCPP at a parameter file containing zeros ($(P-L)\_0$ and $q\_{v0}$) or infinities ($\tau = -1/\frac{\partial (P-L)}{\partial q\_{v}}$) to see if the problem goes away
- Point CCPP to a file where $\tau = -1/\frac{\partial (P-L)}{\partial q\_v}$ is capped at two months so water vapor stays closer to climatology

DWesl:
## Description

Adds arrays to accumulate the water vapor photochemistry terms and write them out.  

This PR should not change model output, but it will change water vapor diagnostics in the stratosphere and mesosphere to reflect changes from methane oxidation and photodissociation.

I should mention that one of the parameters gets called `dtend_qv_ozmix`, because the $-\frac{1}{\tau}(r-r_0)$ term it represents here is analagous to the $\left.\frac{\partial(P-L)}{\partial r}\right|_0 (r - r_0)$ ozone photochemistry diagnostic called `dtend_o3_ozmix`.  I don't know how to change the CCPP machinery to change the name of the diagnostic, nor what to change it to if I did.

### Issue(s) addressed

- fixes ufs-community/ccpp-physics#255

## Testing

- Tested on Hera with default global-workflow compilers, probably Intel
- There should not be changes outside diagnostic output, so existing regression tests should suffice
- I need to look up how to run the regression tests, and have not run them
- This should not require baseline updates, unless they track specific humidity diagnostics from the stratosphere or mesosphere

## Dependencies

- Depends on ufs-community/ccpp-physics#265 for the code to actually calculate the specific humidity tendencies


grantfirl:
@DWesl I don't see a ufs-weather-model pull request for this work. Are you going to add one?

DWesl:
> @DWesl I don't see a ufs-weather-model pull request for this work. Are you going to add one?

Oh, right.  What branch would that be against?  `develop` is pulling in more FV3 changes than mine

grantfirl:
> > @DWesl I don't see a ufs-weather-model pull request for this work. Are you going to add one?
> 
> Oh, right. What branch would that be against? `develop` is pulling in more FV3 changes than mine

It needs to go into develop.

DWesl:
ufs-community/ufs-weather-model#2707 created to pull these changes into that repository.

DusanJovic-NOAA:
Merged via #990

PR #958, opened 2025-05-01
Add switch to disable InfoBroadcast in fv3 cap run phase2

### LINKED ISSUE #955: Add switch to disable InfoBroadcast in fv3 cap run phase2
In the [PR](https://github.com/NOAA-EMC/fv3atm/pull/835) that added support for varying length of accumulation buckets, we added code that synchronizes the field bundle attributes from fcst to write grid components in fv3 cap. [Here](https://github.com/NOAA-EMC/fv3atm/blob/develop/fv3/fv3_cap.F90#L1241-L1254). Unfortunately, this introduced undesirable slow down due to the blocking of entire fv3 cap (call to ESMF_InfoBroadcast works across all fv3 cap PETs). At this moment there is no version of InfoBroadcast that will broadcast data across a subset of fv3 atm PETs, specifically across fcst PETs and only one of the write grid component PETs, (the active one).

Since the slow-down is significant in GFS v17 configurations and since GFS v17 does not use bucket lengths that are changing during the run, it was suggested that we introduce a model configure switch that will skip this block of code. The switch will be false by default, so by default that block of code will not be executed.

DusanJovic-NOAA:
## Description

Add model_configure switch `sync_fcst_info_to_wgc`, false by default, that will skip calling InfoBroadcast in fv3 cap run phase 2. 

### Issue(s) addressed
- fixes noaa-emc/fv3atm/issues/955

## Testing

How were these changes tested?  ufs regression test
What compilers / HPCs was it tested with?  Intel/GNU on Hera
Have the ufs-weather-model regression test been run? Yes.
On what platform? Hera  
Will the code updates change regression test baseline? No.

## Dependencies

No.


fv3/fv3_cap.F90
```diff
@@ -394,6 +395,10 @@ subroutine InitializeAdvertise(gcomp, rc)
       enddo
 
       call ESMF_ConfigGetAttribute(config=CF, value=time_unlimited, label ='time_unlimited:', default=.false., rc=rc)
+      if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, line=__LINE__, file=__FILE__)) return
+
+      call ESMF_ConfigGetAttribute(config=CF, value=sync_fcst_info_to_wgc, label ='sync_fcst_info_to_wgc:', default=.false., rc=rc)
```
junwang-noaa, line 401: Add comments: !! sync_fcst_info_to_wgc: flag to synchronize EMSF_info from fcst grid comp to write grid comps. Needs ESMF update to allow this to be done with write grid component independently.

DusanJovic-NOAA: Added comment in https://github.com/NOAA-EMC/fv3atm/pull/958/commits/27c5b585fdea06bdb9b2dd77d19f7ddc2e3281c2

junwang-noaa (approved):

jkbk2004 (approved):

PR #950, opened 2025-04-17
Combination PR (use FMS tracer_manager for tracer parsing, ECMWF non-stationary GWD changes, other misc. physics changes)

grantfirl:
## Description

This PR contains many physics bugfixes and updates described in https://github.com/ufs-community/ccpp-physics/pull/295. The only physics changes requiring changes to the code in fv3atm are as follows:

- Removal of gfdl_parse_tracers.F90 file from physics: https://github.com/NCAR/ccpp-physics/pull/1087
  - Currently, there is a dependence of GFS_typedefs.F90 on a parse_tracers module in the GFDL MP scheme in the CCPP physics. This was undesirable for NRL NEPTUNE (see https://github.com/NCAR/ccpp-physics/pull/1087). For the SCM, the parse_tracers subroutine was moved to GFS_typedefs.F90 itself, since that made the most sense (see https://github.com/NCAR/ccpp-scm/pull/569). For the FV3, the `get_tracer_index` function already exists within the FMS library and is used elsewhere in the code, so, to avoid confusion, it seems logical for GFS_typedefs.F90 to use the FMS version too. A wrapper for the get_tracer_index was added to GFS_typedefs.F90 because the FMS version returns a large negative integer for a missing tracer, whereas the rest of the physics code expects missing tracers to have an integer value of -99.
- Changes required to work with the ECMWF non-stationary gravity wave drag provided by @BoYang-NOAA: https://github.com/NOAA-EMC/fv3atm/pull/952
  - Add ECMWF Non-stationary gravity wave scheme in UGWDV1 gravity wave suite. If do_ngw_ec is set to true, use the ECMWF NGWD scheme. Conversely, if do_ngw_ec is false, use the original UGWDv1 NGWD scheme. do_ngw_ec is set to false by default.
  - tested on Hera with intel; WCOSS2
- Combines https://github.com/NOAA-EMC/fv3atm/pull/968 (credit @SamuelTrahanNOAA )
  - fixes https://github.com/ufs-community/ccpp-physics/issues/282 

### Issue(s) addressed

None

## Testing

There has been significant testing of individual pieces of this work. Each PR that had been targeted to ufs/dev individually had been tested by itself using UFS RTs on supported platforms. The update to NCAR:main branch was tested together with the removal of gfdl_parse_tracers.F90 changes in GFS_typedefs using UFS RTs on Hera.

## Dependencies

https://github.com/ufs-community/ccpp-physics/pull/295

PR #968, opened 2025-05-15
Bug fix to maxMF check: only access when do_mynnedmf=.true.

### LINKED ISSUE #282: Crash in cu_c3_driver.F90 when not using mynn edmf due to access of unallocated maxMF array.
# Description

When C3 is run without MYNN EDMF, the model can access the unallocated maxMF array and crash. It happens here in `cu_c3_driver.F90`:

```f90
      if((dx(i)<6500.).and.do_mynnedmf.and.(maxMF(i).gt.0.))ierr(i)=555
```

There is no guard against maxMF being accessed when do_mynnedmf is false. That's because Fortran has no short-circuit logical operators. A standard-conforming compilation will evaluate all parts of that "if" condition, even if any of them are false. This is in contrast to nearly all other languages, which would stop at the first false condition in a sequence of logical "and" operations. That's because nearly every language has short-circuit logical operators, but Fortran doesn't.

When you compile with optimization, the optimizer may use short-circuit operators which violate the standard if the compiler decides that it won't affect the results. Nothing in that line involves function calls, so the compiler could safely conclude short-circuit operators are fine if all arrays are allocated. If any array is not allocated, that line should crash. It's reasonable for an optimizer to assume you won't ask it to dereference null pointers.

The correct code is:

```f90
      if(do_mynnedmf) then
        if((dx(i)<6500.).and.(maxMF(i).gt.0.))ierr(i)=555
      endif
```

That code will only access maxMF (and dx) when do_mynnedmf is true.

## Steps to Reproduce

1. Make a custom physics suite which uses C3 but not mynn edmf.
2. Compile that suite in debug mode.
3. Generate compatible input conditions.
4. Run with those input conditions and the new suite. Witness a crash.

## Additional Context

This came up in a project to replace samf with C3 in the HAFS "A" configuration.

SamuelTrahanNOAA:
## Description

Bug fix to https://github.com/ufs-community/ccpp-physics/issues/282 to guard against accessing maxMF in `cu_c3_driver.F90` when that array is not allocated.

See issue for full explanation.

### Issue(s) addressed

- fixes https://github.com/ufs-community/ccpp-physics/issues/282

Experimental physics suite that uses C3 but not mynn edmf crashed without these changes but ran with the changes. See issue for full description.

Hera and Hercules with the Intel compiler..

The experimental physics suite that triggers these bugs doesn't self-reproduce, so I cannot make a regression test for it yet. The reproducibility bug is unrelated and will be the subject of another Issue shortly.

Yes. Hera.

No. Either it crashes or it doesn't.

Done.

## Dependencies

- waiting on ccpp-physics https://github.com/ufs-community/ccpp-physics/pull/283

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,8 +8,8 @@
   branch = main
 [submodule "ccpp/physics"]
   path = ccpp/physics
-  url = https://github.com/ufs-community/ccpp-physics
-  branch = ufs/dev
+  url = https://github.com/SamuelTrahanNOAA/ccpp-physics
+  branch = bugfix/maxmf-check
 [submodule "upp"]
   path = upp
   url = https://github.com/NOAA-EMC/UPP


grantfirl:
@SamuelTrahanNOAA This has been combined into https://github.com/NOAA-EMC/fv3atm/pull/950 for helping with the UFS merge queue workload. This should be left open until that is merged in case the combination PR has problems.

DusanJovic-NOAA:
Merged via #950

PR #971, opened 2025-05-22
Allow photochemistry schemes to have parameters at different times.

DWesl:
## Description

Allow photochemistry schemes to have parameters at different times.
In particular, allow the ozone scheme to have two extra sets of parameters during the ozone hole without crashing the model.

There should be no change in results for forecasts using the old (pre-NOAA-EMC/global-workflow#3386) ozone parameter file.  There will be changes to stratospheric and mesospheric water vapor from the latter half of August to the first half of October, which should bring them in line with the values from before the ozone parameter file changed.
Runs starting after mid-October each year should no longer crash when using the new ozone parameter file.

### Issue(s) addressed

- Might be related to NOAA-EMC/global-workflow#3673, depending on when exactly its forecasts started (1 May runs should be fine, 1 November runs are likely to crash after the first time step)

## Testing

> How were these changes tested?  

Ran a twelve-hour C96 forecast from 2021-11-01T00:00: it crashed after the first time step before this change and got through the first hour after.

> What compilers / HPCs was it tested with?  

global-workflow defaults on Hera

> Are the changes covered by regression tests? (If not, why? Do new tests need to be added?)  

They should be covered by ufs-weather-model tests: I haven't worked out how to run the FV3 tests.

> Have the ufs-weather-model regression test been run? On what platform?  

Currently running on Hera

> - Will the code updates change regression test baseline? If yes, why? Please show the baseline directory below.

It depends on the tests run: if it uses the new ozone parameter file and extends past mid-August, stratospheric and mesospheric water vapor will be affected.  If it extends past mid-October, it should no longer crash.  I'm pretty sure the second case isn't covered, but I'm not sure if any tests use the new ozone parameter file.

> - Please commit the regression test log files in your ufs-weather-model branch

I will once it's done running.

## Dependencies

- waiting on ufs-community/ccpp-physics#286

### DIFF
--- a/ccpp/data/CCPP_typedefs.meta
+++ b/ccpp/data/CCPP_typedefs.meta
@@ -3126,7 +3126,6 @@
   dimensions = (starting_x_direction_index_alloc1:ending_x_direction_index_alloc1,starting_y_direction_index_alloc1:ending_y_direction_index_alloc1,1:vertical_dimension_for_fast_physics,1:number_of_gases_for_multi_gases_physics)
   type = real
   kind = kind_dyn
-  active = (number_of_gases_for_multi_gases_physics > 0)
 [qv]
   standard_name = water_vapor_specific_humidity_at_Lagrangian_surface
   long_name = water vapor specific humidity updated by fast physics at Lagrangian surface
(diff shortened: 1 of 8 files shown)


DWesl:
Merged into #949

PR #981, opened 2025-06-13
Update cloud radiation property calculations and remove Zhao-Car MP schemes

Qingfu-Liu:
## Description

This code updates include the following changes
   1) Merge subroutines for the cloud radiation property calculations
   2) Remove Zhao-Car MP schemes  

### Issue(s) addressed

Reduce the number of subroutines and remove Zhao-Car MP schemes which is no longer used

## Testing

   Results are binary identical after the code changes  
   Will add the regression tests late (have run some of the regression tests on HERA, but not run all the regression tests).

No baseline changes (binary identical)

## Dependencies

No dependence


grantfirl:
@DusanJovic-NOAA This can be closed. It was combined with another PR.

PR #873, opened 2024-10-07
Generalize UFS atmospheric component. Bring in MPAS dycore.

dustinswales:
## Description

This pull request contains changes to generalize atmospheric component of the UFS Weather Model:

1) The reorganization of the repository is to achieve the following goals:
- Move dynamical core specific code into subdirectories.
- Move common code (e.g. physics) into shareable, cross-dycore, higher-level directories. This is outlined in Figure 1 of the ["Tiger Team" report.](https://docs.google.com/document/d/112EcPWssB81dwmlNruLgODxShRa-Ts5Kb8zquoU-i5c/edit?usp=sharing)

2) Refactor the CMakeLists to build an atmospheric component with either an FV3 
or MPAS dynamical core.

3) Add CMakeLists.txt to build MPAS core_atmosphere component. The CMake recipes from the "MPAS atmospheric core" were adapted to work in the UFS build system.

4) Modify CCPP physics CmakeLists.txt to build independent MPAS/FV3 couplings to the physics.

5) Create "stub" mpas infrastructure. This includes stubs for the MPAS atmospheric component driver and overlaying NUOPC cap.

With these changes, the MPAS dynamical core builds within UFS.
Next steps, build atmospheric driver for MPAS, which includes creating new typedefs for coupling between the CCPP and MPAS.

### Issue(s) addressed
N/A

## Testing

Tested on Hera using Intel.
All UFS regression tests pass, see /scratch1/BMC/gmtb/Dustin.Swales/UFS/MPAS/ufs-weather-model/tests/logs/RegressionTests_hera.log
There is a new test in tests/rt.conf to build the MPAS dynamical core (just build, no run)

## Dependencies

This PR introduces a new submodule to the UFS atmosphere, the MPAS-Model. A forked copy of the authoritative MPAS-Model has been created under the ufs-community GitHub organization.

# Requirements before merging
N/A

PR #982, opened 2025-06-18
MPAS dycore + Atmosphere Generalization

dustinswales:
## Description
This pull request introduces the dynamical core from the Model for Prediction Across Scales (MPAS) and contains changes to generalize atmospheric component of the UFS Weather Model to accomodate multiple dynamical cores. 

This PR supersedes #873 

Additionally, a common thread woven through these changes is the migration from the `fv3atm` naming convention to using `ufsatm-fv3` or `ufsatm-mpas`

1) Build system changes:
- Refactor the CMakeLists to build an atmospheric component with either an FV3 or MPAS dynamical core, coupled to the CCPP Physics. Also, add ufsatm-mpas build to CmakeLists.txt
- Introduce a new submodule for the MPAS dynamical core, mpas/MPAS-Model.
- Add CMakeLists.txt to build MPAS core_atmosphere component. The CMake recipes from the "MPAS atmospheric core" were adapted to work in the UFS build system.
- Modify CCPP physics CmakeLists.txt to build independent MPAS/FV3 couplings to the physics.

2) Generalization (physics):
- Partition ccpp/CCPP_Driver.F90 to work with both FV3/MPAS physics-to-dynamics coupling strategies.
- Allow ccpp/CMakeLists.txt to build with either MPAS or FV3 physics. *Note* This may not be necessary once all the physics are hooked up with MPAS.
- Create FV3 and MPAS specific ccpp/config/ccpp_prebuild_config.py files. *Note* This may not be necessary once all the physics are hooked up with MPAS.
- Began partitioning ccpp/data/GFS_typedefs.F90 for FV3 and MPAS. Working through initialization (e.g physics namelist ingestion)
- Create ccpp/driver/MPAS_init.F90 from ccpp/driver/GFS_init.F90

3) Generalization (atmosphere)
- Create common module for code to be shared across dycores in the UWM, ufsatm_util.F90. Initially, a single routine is shared by the MPAS and FV3 atmospheric drivers, needed to set up the tracers/constituents/scalars/WhateverYourHostCallsThese.
- Move io/ out from subdirectory of fv3/. Any dycore specific pieces with be distinguished when we generalize the write-grid  component for MPAS and FV3.

4) Added initial CCPP suite definition file for MPAS enabled RRFS. Only RRTMG initialization working.

5) New typedefs modules for MPAS. *Note* This may be merged with GFS_typedefs.F90 in the future.

6) Create mpas coupling infrastructure for the UWM. This includes ESMF hooks for the MPAS atmospheric driver and unifying the NUOPC cap. ATTN: @danrosen25 
- Create MPAS based atmospheric driver for coupling the dynamical core with the CCPP Physics, mpas/atmos_model.F90
- Create ESMF forecast grid component with appropriate hooks to the atmospheric driver, mpas/module_fcst_grid_comp.F90
- Rename and generalization NUOPC Cap from fv3atm_cap.F90 to ufsatm_cap.F90.
- Add Host controlled Macro to NUOPC Cap to use forecast grid component for selected dynamical core, ATMFV3 and ATMMPAS.
- 
7) Create initial physics-to-dynamics coupling infrastructure, mpas/atmos_coupling.F90
- Define data containers for input/export state between the MPAS dycore and CCPP Physics.
- Create initial routines to translate to-and-from the MPAS and CCPP state variables.

With these changes, the MPAS dynamical core builds and runs within the UWM, albeit it's not producing valid forecasts and much work remains to achieve this

There are no changes to any existing tests.

### Issue(s) addressed
N/A

## Testing
Tested on Hera. 
All existing UWM RTs pass.
No changes to baselines.
New UWM RTs for MPAS.

## Dependencies
N/A


README.md
```diff
@@ -1,25 +1,33 @@
...
 The subcomponents include:
 
- - The Finite-Volume Cubed-Sphere (FV3) dynamical core, originally
-from the [Geophysical Fluid Dynamics
-Laboratory](https://www.gfdl.noaa.gov/).
- - The Common Community Physics Package (CCPP) supported by the
+- The Finite-Volume Cubed-Sphere (FV3) dynamical core, originally
+   from the [Geophysical Fluid Dynamics Laboratory](https://www.gfdl.noaa.gov/).
+
+- The Model for Prediction Across Scales - Atmosphere (MPAS-A)	dynamical
```
climbfuji:
```suggestion
- The Model for Prediction Across Scales - Atmosphere (MPAS-A) dynamical
```

ccpp/CCPP_driver.F90
```diff
@@ -56,14 +56,26 @@ subroutine CCPP_step (step, nblks, ierr)
     character(len=*),         intent(in)  :: step
     integer,                  intent(in)  :: nblks
     integer,                  intent(out) :: ierr
+    character(len=*),         intent(in),optional  :: dynamics
```
climbfuji: I would prefer this to be a mandatory argument, not an optional one.

ccpp/CCPP_driver.F90
```diff
@@ -205,38 +237,61 @@ subroutine CCPP_step (step, nblks, ierr)
...
+           if (trim(dycore)=="mpas") then
+              ! Microphysics
+              call ccpp_physics_run(cdata_block(nb,ntX), suite_name=trim(ccpp_suite), group_name="microphysics", ierr=ierr2)
+              if (ierr2/=0) then
+                 write(0,'(2a,3(a,i4),a)') "An error occurred in ccpp_physics_run for group ", "microphysics", &
+                                           ", block/chunk ", nb, " and thread ", nt, " (ntX=", ntX, "):"
+                 write(0,'(a)') trim(cdata_block(nb,ntX)%errmsg)
+                 ierr = ierr + ierr2
+              endif
+           endif
```
climbfuji, line 298: Should there be an `else` statement that raises an error, since `microphysics` is only valid for `mpas` at the moment?

ccpp/driver/MPAS_init.F90
climbfuji, line 1: There is a lot of duplicate code between this file and GFS_init.F90 ...

mpas/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+# Install Fortran module directory
+install(DIRECTORY ${CMAKE_BINARY_DIR}/${MPAS_MODULE_DIR}/ DESTINATION ${CMAKE_INSTALL_LIBDIR}/${MPAS_MODULE_DIR}/)
+
+###############################################################################
+# Build MPAS libraries...
+###############################################################################
+
+# MPAS Utilities (Externals)
+add_subdirectory(MPAS-Model/src/external/ezxml)
+add_subdirectory(MPAS-Model/src/external/SMIOL)
```
climbfuji: Since you are using SMIOL, why are you looking for PIO in one of the other CMakeLists.txt? SMIOL uses pnetcdf only if I am not mistaken.

climbfuji (commented):
I skipped over the ESMF-NUOPC code changes/additions.

ccpp/CCPP_driver.F90
dustinswales: Agreed. Will change.

ccpp/driver/MPAS_init.F90
dustinswales: Yeah... I will unify the two.

ccpp/CCPP_driver.F90
dustinswales: Good point. I will add an error.

mpas/CMakeLists.txt
dustinswales: We are using PIO, so this should be removed here.

ccpp/CCPP_driver.F90
dustinswales: Done.

ccpp/CCPP_driver.F90
dustinswales: Done.

mpas/CMakeLists.txt
dustinswales: Done.

dustinswales:
@DusanJovic-NOAA @grantfirl Would you mind reviewing this PR whenever you have the time. (Sorry, I can't request reviewers in this repo).

ccpp/CMakeLists.txt
```diff
@@ -4,24 +4,38 @@ project(CCPP-UFS
...
+# MPAS dynamical core
+#------------------------------------------------------------------------------
+if (MPAS)
+  message(STATUS "Build CCPP interface to MPAS dynamical core")
+  set(CCPP_TARGET mpasccpp)
+  set(CCPP_PREBUILD_CONFIG config/ccpp_prebuild_config_mpas.py)
+  set(DYCORE_CCPP_SRCS
+	CCPP_driver.F90
+        )
+  add_definitions(-DFV3)
```
DusanJovic-NOAA: Is this intended, to define FV3 macro when MPAS is used? If it is, it's confusing.

ccpp/data/GFS_typedefs.F90
```diff
@@ -4306,6 +4308,46 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
 !--- NRL ozone physics
     character(len=128) :: err_message
 
+    !--- If initializing model with FV3 dynamical core.
+    if (trim(dycore) == 'FV3') then
```
DusanJovic-NOAA: Here the uppercase character string 'FV3' is used to identify a dycore, while, for example, in CCPP_driver lowercase 'fv3' is used.

```
      if (trim(dycore)=='fv3') then
         ! call timestep_finalize for "phys_ps"---required for Land IAU
```

Can we standardize this to either 'FV3' or 'fv3'?

Another possibility is to define integer parameters somewhere, fe. DYCORE_FV3/DYCORE_MPAS to some values and use them with integer DYCORE_USED variable set to either one and use simple integer comparison instead of 'trimming' character strings everywhere.

climbfuji (commented):
How is ESMF handled in MPAS in this version?

By default, MPAS uses its own, stripped-down version of ESMF that provides just the clock routines. But you can tell MPAS to use an external ESMF instead, which makes sense in the context of the UFS that requires and configures its use in the CMake build system.

ccpp/CMakeLists.txt
DusanJovic-NOAA: I also see ATMFV3 macro defined in the top-level CMakeLists.txt (`add_definitions(-DATMFV3)`). Can we use the same macro throughout this repo to identify fv3 or mpas atm/dycores?

ccpp/data/GFS_typedefs.F90
DusanJovic-NOAA: Another advantage of using integer constants is for example if I mistype 'fv3' here as 'fv4' code will compile fine, without any warnings, and just (silently) skip this if block, while if I mistype `if(DYCORE_USED == DYCORE_FV4)` it will fail to compile, which is better.

Also, 'grepping' for DYCORE_FV3 is easier than fv3.

ccpp/CMakeLists.txt
grantfirl: Agreed. If the MPAS build needs something that gets built because of the FV3 definition, it would be better to split that out and call it independently when -DMPAS or -DFV3 are set.

ccpp/config/ccpp_prebuild_config_mpas.py
```diff
@@ -0,0 +1,159 @@
...
+# Auto-generated makefile/cmakefile snippets that contain all caps
+CAPS_MAKEFILE   = '{build_dir}/physics/CCPP_CAPS.mk'
+CAPS_CMAKEFILE  = '{build_dir}/physics/CCPP_CAPS.cmake'
+CAPS_SOURCEFILE = '{build_dir}/physics/CCPP_CAPS.sh'
+
+# Directory where to put all auto-generated physics caps
+CAPS_DIR = '{build_dir}/physics'
+
+# Directory where the suite definition files are stored
+SUITES_DIR = '../ccpp/suites'
```
grantfirl, line 144: Are suites going to be shared verbatim between MPAS and FV3? If we are going to need host-specific interstitials (as prophesied in the past), it may be helpful to have separate directories in ccpp/suites for each dycore?

ccpp/config/ccpp_prebuild_config_mpas.py
```diff
@@ -0,0 +1,159 @@
...
+STATIC_API_DIR = '{build_dir}/physics'
+STATIC_API_CMAKEFILE = '{build_dir}/physics/CCPP_STATIC_API.cmake'
+STATIC_API_SOURCEFILE = '{build_dir}/physics/CCPP_STATIC_API.sh'
+
+# Directory for writing HTML pages generated from metadata files
+# used by metadata2html.py for generating scientific documentation
+METADATA_HTML_OUTPUT_DIR = '{build_dir}/physics/physics/docs'
+
+# HTML document containing the model-defined CCPP variables
+HTML_VARTABLE_FILE = '{build_dir}/physics/CCPP_VARIABLES_FV3.html'
```
grantfirl: references FV3 instead of MPAS

ccpp/config/ccpp_prebuild_config_mpas.py
```diff
@@ -0,0 +1,159 @@
...
+
+# LaTeX document containing the provided vs requested CCPP variables
+LATEX_VARTABLE_FILE = '{build_dir}/framework/doc/DevelopersGuide/CCPP_VARIABLES_FV3.tex'
```
grantfirl: Ditto

ccpp/data/GFS_typedefs.F90
```diff
@@ -4306,6 +4308,46 @@ subroutine control_initialize (Model, nlunit, fn_nml, me, master, &
...
+       if (.not. present(tile_num)) then
+          write(6,*) 'ERROR: <tile_num> is required when using FV3 dynamical core'
```
grantfirl, line 4318: Aside from writing out a message, should there be an error flag that gets passed back to the calling routine so that it can stop gracefully? Also, return immediately on error?

ccpp/data/GFS_typedefs.F90
grantfirl: In addition, why not make `dycore` (whether it remains a string or gets changed to an integer) part of the GFS_control DDT so that it doesn't need to passed around in addition to the DDT?

dustinswales:
> How is ESMF handled in MPAS in this version?
> 
> By default, MPAS uses its own, stripped-down version of ESMF that provides just the clock routines. But you can tell MPAS to use an external ESMF instead, which makes sense in the context of the UFS that requires and configures its use in the CMake build system.

We are using the UFS (external) ESMF in MPAS. In mpas/CmakLists.txt it reverts back to MPAS's internal ESMF if one is not provided by the host.

ccpp/CMakeLists.txt
dustinswales: Sorry, I will do a "macro" cleanup.

ccpp/data/GFS_typedefs.F90
dustinswales: All good ideas. I've included them. Thanks.

ccpp/config/ccpp_prebuild_config_mpas.py
dustinswales: Because of the different coupling between physics/dynamics in FV3/MPAS, the suites will need to have different groupings.
Where the SDFs live and what they are called is a good question!

Currently the FV3 SDFs are using the following suite naming convention (outside the RRFS ensemble):
ccpp/suites/DYCORE_UFSAPP_PHYSICS.xml

Of course I didn't follow that here for RRFS MPAS, instead I used 
ccpp/suites/UFSAPP_DYCORE_PHYSICS.xml. 

If we go with distinct suite directories for each dycore, we would have something like ccpp/suites/DYCORE/UFSAPP_PHYSICS.xml

I'm indifferent to which we decide. There are instances in this PR where I partitioned things into dycore directories, and there are instances where the MPAS and FV3 pieces exist next to one another. For UFSATM/ccpp it is more of the latter, and this was because I didn't know how much would be shareable. I "think" that a lot will be shareable, but certainly not the SDFs

dustinswales:
@DusanJovic-NOAA @grantfirl 
Thanks for the feedback.
I did a "CMake macro cleanup" and switched to using an integer identifier for the selected dycore.

DusanJovic-NOAA:
@dustinswales I started seeing these warnings about deprecated cmake version from one of the CMakeLists.txt recently:

```
CMake Deprecation Warning at FV3/ccpp/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```
Since you are already updating various CMakeLists.txt files in this PR, can you please update the version to, for example, 3.19 which is the version currently required in the top-level CMakeLists.txt.

grantfirl (approved):

climbfuji (commented):
This looks good to me. I'll wait for the conflict resolution and regression testing before I approve. Thanks for addressing all my comments and that of the other reviewers.

DusanJovic-NOAA (approved):

ccpp/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ if (FV3)
         driver/GFS_init.F90
         CCPP_driver.F90
         )
+
```
dustinswales, line 24: @DusanJovic-NOAA I had to add this line back in. 
Without this, the fast-physics called from FV3 will try to open a new parallel region in [fv_sat_adj.F90](https://github.com/ufs-community/ccpp-physics/blob/ufs/dev/physics/MP/GFDL/fv_sat_adj.F90#L324), which causes failure for the suites using GFDL MP w/ fast-physics, mostly HAFS configurations.

dustinswales:
@climbfuji @grantfirl @DusanJovic-NOAA 
I had to revert one change to the ccpp/CMakeLists, which was causing failures in the configurations using the GFDLMP w/ fast-physics.
Plus one other small change within MPAS to remove the timestamp from the log file. This makes the log reproducible.
All the UWM tests are passing, re-running now and will update logs in UWM PR when complete.

jkbk2004 (approved):

PR #1001, opened 2025-08-19
Replace icplocn2atm (integer) with use_oceanuv (logical)

### LINKED ISSUE #2704: setting icplocn2atm should be renamed to clearly indicate its purpose
## Description

The logical ``cplocn2atm`` is default true and exists to turn on ocn->atm coupling of SST. A second integer setting was later added ``icplocn2atm`` which then enables coupling surface currents from ocn->atm. The default for ``icplocn2atm`` is set 0, which is to *not* use ocean currents.

This is confusing. Coupling ocean currents is optional in the same way that ``use_cice_alb`` is. When coupled to CICE (``cplice``), certain fields are essentially mandatory (in the same way SST is for the ocean coupling) but then optionally CICE albedos can also be used (whether this should in fact be optional is another issue, since CICE albedos correctly account for the ice thickness categories in creating a composite ice albedo. This means the impact of very thin ice on the albedo is included vs using parameterized ice albedo.) 

The setting ``icplocn2atm`` doesn't actually indicate what a user is optionally including. It is not an integer analog of ``cplocn2atm`` as the name implies. Additionally, there is a request to allow the use of ocean salinity. What are we going to call that--``jcplocn2atm``?

We should rename ``icplocn2atm`` to a) be a logical as the other coupling options are and b) renamed to ``use_oceanuv``. When adding ocean salinity, the option would be ``use_ocean_salinity``. 

Related to #2676

DeniseWorthen:
## Description

Replaces the current integer option of icplocn2atm with a logical flag use_oceanuv

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/2704

## Testing

The only current UFS RT which utilizes the ocean surface currents in the ATM is B4B with this change.

## Dependencies

- waiting on https://github.com/ufs-community/ccpp-physics/pull/307
- waiting on https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/389
- required by https://github.com/ufs-community/ufs-weather-model/pull/2867


DusanJovic-NOAA (approved):

grantfirl:
https://github.com/ufs-community/ccpp-physics/pull/307 is merged, so this is ready to revert .gitmodules and update ccpp/physics

jkbk2004 (approved):

PR #1051, opened 2026-01-05
Remove Zhao-Carr cloud schemes and combine Xu-Randall cloud fraction

Qingfu-Liu:
## Description

This PR is associated with ccpp-physics PR#337.  The changes are:

1. Remove Zhao-Carr cloud schemes which are no longer used in the CCPP physics suites. The new code changes remove cloud calculation for "progcld_zhao_carr" and "progcld_zhao_carr_pdf".
2. Create a new function cld_frac_XuRandall in the module_radiation_clouds, and replaced the subroutines "cloud_fraction_mass_flx_2", "cloud_fraction_mass_flx_1", and "cloud_fraction_XuRandall" with the new function.
3. remove the function cld_frac_XuRandall from program GFS_rrtmgp_cloud_mp.F90, and replaced it with the newly created function cld_frac_XuRandall in the module_radiation_clouds.

Tests Conducted:

Regression tests has been conducted using the default suites, and the results are binary identical.

### Issue(s) addressed

Remove Zhao-Carr cloud schemes which are no longer used in the CCPP suites, and combine the Xu-Randall cloud fraction calculations

## Testing

Regression tests needed for all platforms. I have run the regression tests on Ursa, and the tests show no results changed
No other tests needed.

## Dependencies

Need merge the ccpp-physics PR#337 first.


gspetro-NOAA:
Could we get approvals on this PR so that we can process https://github.com/ufs-community/ufs-weather-model/pull/3042?

dustinswales (approved):
Looks good to me. 
(However, I didn't go through all of the physics to see that all the references to ZC have been removed.)

gspetro-NOAA:
Ideally, we need one more review on this PR before we can schedule https://github.com/ufs-community/ufs-weather-model/pull/3042

grantfirl (approved):
Looks good to me. Someone in charge of UFSATM should probably review whether it is OK to remove the suites in the `suite_not_used` directory in this PR. By definition, these are not used, I suppose, and since the ZC MP scheme is being removed, these particular suites would need to have MP replaced with something else, in which case, they're not really the same suites anymore. Given this, I think it is completely prudent to remove them, but just want to point this out.

DusanJovic-NOAA:
> Looks good to me. Someone in charge of UFSATM should probably review whether it is OK to remove the suites in the `suite_not_used` directory in this PR. By definition, these are not used, I suppose, and since the ZC MP scheme is being removed, these particular suites would need to have MP replaced with something else, in which case, they're not really the same suites anymore. Given this, I think it is completely prudent to remove them, but just want to point this out.

In fact we should remove the entire 'suites_not_used' directory. Maybe not in this PR but in general. It's been almost 3 years now since we moved unused SDFs into it, this location was only meant to be temporary in case somebody needed to use one of those SDFs. I am not aware that anybody actually used any of those since then.

DusanJovic-NOAA (approved):

gspetro-NOAA:
This PR can be merged anytime (as soon as repo checks have passed). Testing completed successfully on https://github.com/ufs-community/ufs-weather-model/pull/3042, and CCPP PR 337 has been merged. `.gitmodules` and the CCPP hash have been updated.

PR #1079, opened 2026-03-20
Some more MPAS functionality 

dustinswales:
## Description
Contains updates to the UFS interface to the MPAS dynamical core. 

New advancements include: 
- Constituent handling and scalar advection. New module, ufs_mpas_constituents, which contains interface code between MPAS constituents and the UFS.
- MPAS native output stream. Contained within the _ufs_mpas_io_ module. This module also includes procedures for reading in the various MPAS input streams (e.g., invariant, input, sfc_input)
- Support for lateral boundary condition (LBC) stream. New module with code related to regional MPAS configurations, _ufs_mpas_boundaries_. The LBCs are updated during forecast integration, _run phase in ufs_mpas_subdriver.F90.
- Progress on CCPP physics coupling. New MPAS input stream (sfc_input) for gridded fields needed by the physics. Additional P2D coupling procedures to handle the state evolution between physics group and dycore calls. Added diagnostics to output stream.

### Issue(s) addressed
- [UWM #2782](https://github.com/ufs-community/ufs-weather-model/issues/2782)
- [UWM #2879](https://github.com/ufs-community/ufs-weather-model/issues/2879)

## Testing
On Ursa and Derecho with GNU.
Expanded the existing (barebones) MPAS UWM RTs to a) run 12-hour forecast w/ 3-hour output in RELEASE mode and b) added a new DEBUG test that runs a 3-hour forecast with hourly output.

## Dependencies
https://github.com/ufs-community/ccpp-physics/pull/365


gspetro-NOAA:
Could we get reviews on this PR so that we can process https://github.com/ufs-community/ufs-weather-model/pull/3156?

grantfirl:
@dustinswales The PR description doesn't mention that this is supposed to go in with https://github.com/ufs-community/ccpp-physics/pull/365. Could you fix this?

mpas/atmos_coupling.F90
```diff
@@ -183,64 +205,175 @@ subroutine ufs_mpas_to_physics(physics_state)
...
     ! [k, i] -> [i, k]
-    ! bottom-up -> top-down ordering convention
-    do iCell = 1, nCellsSolve
-       iCol = indicesGlobal(iCell)
-       physics_state % pgr(iCol)    = MPAS_state % pintdry(1,iCell)
-       physics_state % prsl(iCol,:) = MPAS_state % pmiddry(nVertLevels:1:-1,iCell)
-       physics_state % prsi(iCol,:) = MPAS_state % pintdry(nVertLevels+1:1:-1,iCell)
+    ! Retain bottom-up convention
+    do iCol = 1, nCellsSolve
+       physics_state % pgr(iCol)    = MPAS_state % pintdry(nVertLevels+1,iCol)
```
NickSzapiro-NOAA, line 267: Should this be pintdry(1,iCol)? If MPAS vertical levels are bottom-up

DusanJovic-NOAA:
I think the code in the 'mpas' directory should not depend on FMS. FMS is a framework/library used by the FV3 dycore. The code in the 'mpas' directory connects the MPAS dycore with the CCPP and NUOPC driver, none of these components depends on FMS. NUOPC driver uses the ESMF framework and MPAS uses its own internal framework code.

From what I can see by just looking at the code, FMS routines are mostly used for error logging and timing. I think we should either use the ESMF or MPAS equivalent routines for logging and timing, and avoid dependency on FMS.

This does not need to be addressed in the current PR, but I wanted to flag it as a potential future task.

mpas/ufs_mpas_boundaries.F90
NickSzapiro-NOAA, line 1: What's different about this than src/core_atmosphere/dynamics/mpas_atm_boundaries.F ?

mpas/atmos_coupling.F90
dustinswales: Correct.
There is [another development branch](https://github.com/dustinswales/fv3atm/blob/feature/phys2dyn/mpas/atmos_coupling.F90#L271) wading in the wake of this one that begins to tackle the MPAS-to-CCPP coupling (well just radiation at the moment).

dustinswales:
> I think the code in the 'mpas' directory should not depend on FMS. FMS is a framework/library used by the FV3 dycore. The code in the 'mpas' directory connects the MPAS dycore with the CCPP and NUOPC driver, none of these components depends on FMS. NUOPC driver uses the ESMF framework and MPAS uses its own internal framework code.
> 
> From what I can see by just looking at the code, FMS routines are mostly used for error logging and timing. I think we should either use the ESMF or MPAS equivalent routines for logging and timing, and avoid dependency on FMS.
> 
> This does not need to be addressed in the current PR, but I wanted to flag it as a potential future task.

I completely agree. I [opened an Issue](https://github.com/ufs-community/ufs-weather-model/issues/3181) and will include this in the next PR. (FYI. There is a second wave of development that I will open a PR for once this one gets merged.)

mpas/ufs_mpas_boundaries.F90
dustinswales: This module has two routines from src/core_atmosphere/dynamics/mpas_atm_boundaries.F that needed changes for UWM. 
(1) In the UWM we don't use the MPAS stream manager to ingest the LBCs (or any other data). So we have a version of ufs_mpas_atm_update_bdy_tend that references our stream manager, dyn_mpas_read_write_stream.
(2) MPAS has a LBC configuration check, but this also relied on the stream manager.

NickSzapiro-NOAA: Thanks @dustinswales 

I don't know if this is the best place to keep asking questions, but why have different stream managers?
UFS hooks into native options for some other components, like CICE

dustinswales: Seems like a good landing spot for discussion.
There are a few reasons for this, but I should clarify that while we don't use the MPAS Stream manager, we still use the native MPAS routines that the stream manager invokes. Namely mpas_readstream and mpas_writestream.

For MPAS dycore in the UFS, I adopted the I/O developed by NCAR-MMM and NCAR-CGD for CAM-SIMA. In this approach the host-model controls all of the I/O, and consequently all the functionality in MPAS stream managers is not really needed. 

We could be more MPAS-A like for I/O, but we have another effort to create the ESMF write grid component. If it turns out that PIO is satisfactory for operational throughput, we may stick with native MPAS I/O.

NickSzapiro-NOAA: I don't know the best way but hesitate about having different options to do same things. Maybe the workflow folks have preferences (?)

There was also discussion for more flexible output_interval in MPAS too, so it seems there's room for making common ground

It seems hard to replace the ability of write grid component to change output resolution online

dustinswales:
@DusanJovic-NOAA @NickSzapiro-NOAA Is there anything else I need to address? 
I have a nearly completed removing all FMS dependencies, but I was planning on submitting a new PR with those changes. Unless you want me to bring them into here?

DusanJovic-NOAA:
> @DusanJovic-NOAA @NickSzapiro-NOAA Is there anything else I need to address? I have a nearly completed removing all FMS dependencies, but I was planning on submitting a new PR with those changes. Unless you want me to bring them into here?

I think removing FMS dependencies should be a separate PR. I would leave this PR as it is. Just keep it synced with develop.

NickSzapiro-NOAA:
> Is there anything else I need to address?

I haven't looked at the new baselines yet but wonder if it's worth comparing real or idealized test cases to MPAS-A standalone (if that's the reference implementation), thanks to the added transport and output.

DusanJovic-NOAA (approved):

dustinswales:
> > Is there anything else I need to address?
> 
> I haven't looked at the new baselines yet but wonder if it's worth comparing real or idealized test cases to MPAS-A standalone (if that's the reference implementation), thanks to the added transport and output.

@NickSzapiro-NOAA I've looked at the baselines (e.g., T, u, v, q, tracers) and they seem realistic. For the next PR, when the physics is included and we have more diagnostics, I will include some figures from offline and inline MPAS.

ccpp/data/GFS_typedefs.F90
```diff
@@ -6879,7 +6898,7 @@ subroutine control_print(Model)
     class(GFS_control_type) :: Model
 
 !--- local variables
-    integer :: i
+    integer :: i, mpierr
```
NickSzapiro-NOAA, line 6901: I think mpierr is unused

mpas/ufs_mpas_io.F90
```diff
@@ -0,0 +1,2020 @@
...
+ !>
+ !>
+ !> ########################################################################################
+ subroutine read_stream(stream, when, whence, actualWhen, nRecord, ierr)
+   use mpas_io_streams,     only : MPAS_readStream, MPAS_streamTime
+   use mpas_derived_types,  only : mpas_pool_type, mpas_stream_noerr, mpas_stream_type
+
+   type(mpas_stream_type), pointer, intent(inout) :: stream
+   character (len=*), intent(in) :: when
+   integer, intent(in) :: whence
```
NickSzapiro-NOAA, line 905: Are when and whence unused?

It's not clear to me what when/whence/actualWhen mean

ufsatm_cap.F90
```diff
@@ -272,6 +273,7 @@ subroutine InitializeAdvertise(gcomp, rc)
     integer                                :: ngrids
     type(ESMF_Grid)                        :: src_grid, dst_grid
     type(ESMF_Field), allocatable          :: dst_field_mask(:)
+    integer                                :: ierr
```
NickSzapiro-NOAA, line 276: ierr is unused

ccpp/data/GFS_typedefs.F90
dustinswales: Removed.

ufsatm_cap.F90
dustinswales: Removed.

mpas/ufs_mpas_io.F90
dustinswales: Good catch. These are unused I removed.
These are timing arguments that I was testing out when adding the LBC stream.

dustinswales:
@NickSzapiro-NOAA Thanks for the review. I addressed all of your comments.

dustinswales:
@NickSzapiro-NOAA @DusanJovic-NOAA Here are some simple difference plots (Standalone MPAS - Inline (UFS) MPAS) of the state variables at the lowest model level.
Screenshot 2026-04-22 at 11 09 15 AM

DeniseWorthen:
@dustinswales These differences must be at least at the round-off level, surely?

dustinswales:
> @dustinswales These differences must be at least at the round-off level, surely?

@DeniseWorthen I'm sure there is, but I was shocked that they looked this close just eye-balling it.
I'm going to dive a bit deeper and create better figures. Stay tuned...

NickSzapiro-NOAA:
If it's just advection and output, shouldn't MPAS-A standalone match MPAS-A dycore in UFS **exactly**?

dustinswales:
> If it's just advection and output, shouldn't MPAS-A standalone match MPAS-A dycore in UFS **exactly**?

@NickSzapiro-NOAA My expectation is for initial agreement that diverges with forecast time. There are many differences between the modeling systems, so there's that piece of it. But I will also admit that I'm also not sure how a simulation w/o physics should evolve?

Here's a joint PDFs of global mean differences as a function of forecast time and model layer. For all fields there is initial agreement with subsequent drifting apart of the forecasts, especially aloft.

Theta:
Screenshot 2026-04-23 at 11 27 25 AM

qv:
Screenshot 2026-04-23 at 11 28 31 AM

uwind:
Screenshot 2026-04-23 at 11 28 48 AM

vwind:
Screenshot 2026-04-23 at 11 29 01 AM

NickSzapiro-NOAA:
Maybe I'm wrong to assume that the dynamics should match but the physics will differ ... but it looks like differences start near the model top

Is it possible for you to share (here or separately) what you're running for this comparison @dustinswales ?
Is MPAS standalone using the regression test inputs too?

dustinswales:
@NickSzapiro-NOAA @DusanJovic-NOAA I think I've gotten to the bottom of it (Some default nml settings were not correct). 

Here are the jointPDFs after correcting the configuration:
inlineMPAS - standaloneMPAS (theta)
Screenshot 2026-04-24 at 2 58 32 PM

inlineMPAS - standaloneMPAS (v_wind)
Screenshot 2026-04-24 at 2 58 43 PM

inlineMPAS - standaloneMPAS (u_wind)
Screenshot 2026-04-24 at 2 58 55 PM

For comparison, here are the joint PDFs of standaloneMPAS(gnu) - standaloneMPAS(ifx):
(theta)
Screenshot 2026-04-24 at 3 01 25 PM

(v_wind)
Screenshot 2026-04-24 at 3 01 48 PM

(u_wind)
Screenshot 2026-04-24 at 3 02 07 PM


**_So inlineMPAS introduces less differences wrt to standaloneMPAS than does changing the compiler_**!

gspetro-NOAA:
@dustinswales Could you revert .gitmodules and update the CCPP hash to [6904ad4](https://github.com/ufs-community/ccpp-physics/commit/6904ad4b5563ca1a63203993b35fa68de794340f)?

gspetro-NOAA:
This PR can be merged :)

dpsarmie (approved):

Issue #1093, opened 2026-04-13
model hangs if set cpl_imp_dbg=true for cases when ATM runs threaded


DeniseWorthen:
## Description

In ``atmos_model.F90``, there is an optional diagnostic block which writes out merged input fields when ``cpl_imp_dbg=.true.`` 

https://github.com/NOAA-EMC/ufsatm/blob/972a45054d579ff6f0c6cdb751452d29b25c133f/fv3/atmos_model.F90#L3185-L3201

These are used by eg the HAFS regional configs which merge the input fields when non-overlapping domains are present ( ``cpl_imp_mrg=.true.``). The dbg logical is supposed to write out the merged fields for checking. 

Adding the cpl_imp_dbg to the ``input_regional_hafs.nml.IN``
```
diff --git a/tests/parm/input_regional_hafs.nml.IN b/tests/parm/input_regional_hafs.nml.IN
index 7d028cb0..2212a928 100644
--- a/tests/parm/input_regional_hafs.nml.IN
+++ b/tests/parm/input_regional_hafs.nml.IN
@@ -227,6 +227,7 @@
        cplwav         = @[CPLWAV]
        cplwav2atm     = @[CPLWAV2ATM]
        cpl_imp_mrg    = @[CPL_IMP_MRG]
+       cpl_imp_dbg    = .true.
        psautco        = 0.0008, 0.0005
        prautco        = 0.00015, 0.00015
        iau_delthrs    = 6
```

works for all couple HAFS configurations except for the ``hafs_regional_storm_following_1nest_atm_ocn_wav_mom6_intel``, which runs with 2 threads. In that case the model hangs.

### To Reproduce:
Edit the above nml and run the applicable test.

I did see that there is a missing directive for the FBH model imports; Those all have 
-!$omp parallel do default(shared) private(i,j,nb,ix)

but I believe they should be 
-!$omp parallel do default(shared) private(i,j,nb,ix,im)

Fixing this did not resolve the hang.

DeniseWorthen:
I added a simple print statement
```
@@ -4701,6 +4701,7 @@ module GFS_typedefs
     Model%cpl_imp_dbg      = cpl_imp_dbg
     Model%use_med_flux     = use_med_flux

+    print *, 'Rank ', me, ': XX5 - cpl_imp_mrg=', Model%cpl_imp_mrg, ' cpl_imp_dbg=', Model%cpl_imp_dbg
 !--- cdeps inline parameters
```

[...]

Issue #1096, opened 2026-04-14
move cpl_imp_mrg and cpl_imp_dbg to atmos_model_nml


DeniseWorthen:
## Description
The cpl_imp_mrg and cpl_imp_dbg logicals are used only in assign_importata. They are not used by physics but currently reside in the ``gfs_physics_nml``. 

## Solution

Move these logicals to the ``atmos_model_nml``. See #1093 for additional info.

DeniseWorthen:
@BinLiu-NOAA I wanted to make you aware of this issue; I'm planning to do that as part of a followup PR to UFSATM.

BinLiu-NOAA:
Thanks, @DeniseWorthen! This makes sense and sounds reasonable to me. Thanks for making these updates!

PR #1098, opened 2026-04-20
Refactor assign_importdata

### LINKED ISSUE #1085: Refactor assign_importdata to clean up boilerplate and allow unit-testing
## Description
The assign_importdata routine in ``atmos_model.F90`` is full of boilerplate code and 'special use' cases. The routine should be re-factored for clarity, testability and future accommodation with the MPAS (where import fields, ie the ``datar8`` pointer) will presumably have a reduced dimensionality compared to FV3.

It would also be very useful to have a way of writing out a field bundle of the fields being passed to the physics after the importdata manipulation. We have multiple ways of looking at what is sent and received in the importstate fields, but that is not always what is sent to physics. 

One idea would be to create a diagnostic FB in importdata, and sequentially add fields for each imported field that is modified which would point to the values passed to physics rather than just the imported values. This would replace the code block currently inside of the ``cpl_imp_dbg``, which right now writes out individual files for every import field (and which needs modification to work w/ multi-tiled grids anyway).

## Solution
Replace the repeated loops over ``jsc,jec``,``isc,iec`` with a subroutine (``copy2block``) with optional arguments to cover various use cases. Add a unit test for the new subroutine. 

It seems possible to allow this ``copy2block`` to exercise an optional abort or message when an imported value is being replaced w/ an override value (eg, the current use of the "if > 150.0"). 

Add an optional interface (``merge_importfield``) to cover the existing ``GFS_control%cpl_imp_mrg`` use case.
Add a unit test for the new interface.

I have a working [feature branch](https://github.com/DeniseWorthen/ufsatm/tree/feature/copy2block) implementing the copy2block and merge_importfield. This is WIP. Additional work and modifications are still needed.

The copy2block unit test is written by AI (with guidance) and is located in the tests directory (test_module_copy2block). In order to build the unit-test outside of the CI framework, I had to temporarily insert ifdef statements to remove STOCHY_PHYS from the build. To build and run the unit test:

```
module purge
module use ../../modulefiles
module load ufs_ursa.gnu
export CPATH=/usr/lib/gcc/x86_64-redhat-linux/11/include${CPATH:+:$CPATH}
cmake -S . -B build_copy2block -DCMAKE_BUILD_TYPE=Debug -DFV3=ON -DSTOCH_PHYS=OFF -DBUILD_CI_TESTING=ON -DCMAKE_MODULE_PATH="$PWD/fv3/atmos_cubed_sphere/cmake;$PWD/mpas/MPAS-Model/cmake/Modules"
cmake --build build_copy2block --target test_module_copy2block -j 8
./build_copy2block/tests/test_module_copy2block
```

## Alternatives
If applicable, add a description of any alternative solutions or features you've considered.

## Testing:

1. Have you tested the code changes? On what platforms?
2. Have you run regression test in ufs-weather-model or ufs-s2s-model with code changes?
- Will the baseline results change?
- If yes, please give brief explanation of your code updates on the regression test.

## Dependent PRs:

Directly reference any issues or PRs in this or other repositories that this is related to, and describe how they are related. Example:
- required to support noaa-emc/GFDL_atmos_cubed_sphere/issues/<issue_number>
- ncar/ccpp-physics/pull/<pr_number>
- associated ufs-weather-model/pull/<pr_number>

### LINKED ISSUE #1097: Allow unit tests to build and run outside of CI
## Description
For adding and testing new unit tests, it is convenient to do this outside of CI. This allows a quick debug/development cycle. 

## Solution
Move the CMakeModules and stochastic_physics setup logic from the CI (GCC.yml) to the CMake config. 

## Testing:

A solution was provided by @AlexanderRichert-NOAA and has been included in the work for #1085. The following will build and run the UFSATM unit tests. The only caveat is that I've needed to have the GNU UWM module files loaded on a particular platform first so I found it easiest to do this from a checkout of ufs-weather-model. 

```
cd UFSATM
mkdir build
cd build
cmake .. -DFV3=ON -DBUILD_UFSATM_TESTING=ON
make -j8
make test
```

For a quick debug cycle, the following can be used to build and run a single test from within the ``build`` directory:

```
make single_test_name
 ./tests/single_test_name
```

DeniseWorthen:
## Description

* Cleans up boilerplate in ``assign_importdata`` by replacing repeated loops with a call to a new subroutine ``copy2block``. 
* Adds a new interface to be exercised when ``cpl_imp_mrg`` is set true. This feature is used by HAFS to handle the case where the imported fields are not fully overlapped with the ATM domain.
* Adds a debugging FB which can be used to check values sent to physics when imported field values are loaded into the physics arrays but then manipulated prior to exiting the routine. This feature replaces an existing capability (when ``cpl_imp_dbg = .true``) in a more flexible manner.
* Sets the timestring on the file name used when ``DumpFields = .true.`` to be consistent with CMEPS mediator history file naming; that is, all file names are the ending time of the ModelAdvance. This allows the debug FB to also be written with a consistent file naming.
* Adds unit tests for the ``copy2block`` and ``merge_importfield`` routines. Note the unit tests structure was primarily written by AI, with guidance.
* Adds the capability to build and run the UFSATM unit tests outside of CI.

Currently in the develop branch, the existing ``cpl_imp_dbg=.true`` feature hangs for the HAFS-MOM6 case (see #1093). The same is true of the new implementation. The hang is at either the ``FieldWrite`` call in the current develop branch or at the ``FieldBundleWrite`` call in the new implementation. Note that the ``cpl_imp_dbg`` feature was originally added prior to the addition of the HAFS-MOM6 case. 

The debug FB has been tested and works for other regional configurations (ie, hafs atm-wav, regional FBH) and global coupled cases. 

### Issue(s) addressed

- fixes #1085 
- fixes #1097 

## Testing

Testing in the associated UWM feature branch is B4B on hercules for all tests.


DeniseWorthen:
There is a lot of code to review in this PR, so there are several other changes which I have not made at this time, which I think should in fact be made. 

There were also some minor whitespace formatting changes which were unintentional (indenting). I can try to remove those if desired. 

1) I don't think ``lcpl_fice`` is required. Also, I think it better to enclose the loading of CICE6 imports w/in ``if cplice .and. cplflx ...) guards.
2) I don't think the guard on ``dataptr(isc,jsc)`` should be retained
3) I don't think the setting of physics fields to ``-99999.0`` ("over open water - should not be used in ATM") should be retained.
4) I think the cpl_imp_dbg and cpl_imp_mrg flags should be moved to the ``atmos_model_nml`` and removed from the ``gfs_physics_nml`` (#1096).

DeniseWorthen:
@AlexanderRichert-NOAA Could you take a look at why the CI is failing?

AlexanderRichert-NOAA:
Can you try removing the "export <CC/CXX/FC>=mpi<...>" lines in the doc build section? If that doesn't work then I can investigate further.

AlexanderRichert-NOAA:
Nevermind about removing those.

I'm able to get around that particular issue by apt install'ing mpich to make those wrappers available. My best guess is that mpich is no longer being built through Spack in the workflow where the cached Spack installation is generated.

That said, it might make maintenance easier to consolidate the doxygen-generation CI into GCC.yml since there's a fair bit of redundant CI code. Alternatively or in addition, you could add the option to make the documentation a standalone CMake build so you don't have to deal with dependencies etc. (apart from Doxygen itself).

DeniseWorthen:
@AlexanderRichert-NOAA I was able to get enough help from AI to fix the CI build. Right now it is in a throwaway branch  in my fork (https://github.com/DeniseWorthen/ufsatm/tree/test). It moved the doc build into GCC.yml, removed docs.yml and then fixed a "lto1 (“multiple prevailing defs”), which is an LTO-stage failure". 

Would you take a look at the workflow changes in my branch and see if you agree on how it was done?

AlexanderRichert-NOAA:
Yeah that all looks pretty reasonable, I agree it makes sense to move the documentation generation alongside the code coverage rather than being a separate workflow. The only small comment is for clarity it might make sense to not put those no-lto/no-strict-aliasing flags in the "gcov_flags" variable; they could either always be part of the CMake line, or a separate variable that's set unconditionally (not in the debuild=ON/OFF block) so that it's clear what they're for.

ci/CMakeLists.txt
NickSzapiro-NOAA, line 1: We don't update CMakeModules and stochastic_physics much in ufs-weather-model, but I don't know that we'll remember to change them here in ufsatm/ci when we do

AI suggests that you can get the sha hash from ufs-weather-model/develop instead of hardcoding here via like https://api.github.com/repos/ufs-community/ufs-weather-model/contents/CMakeModules?ref=develop
Something like:
```
function(get_ufs_submodule_hash SUBMODULE_NAME OUTPUT_VAR)
    message(STATUS "Querying ufs-weather-model for ${SUBMODULE_NAME} hash...")
    file(DOWNLOAD 
        "https://api.github.com/repos/ufs-community/ufs-weather-model/contents/${SUBMODULE_NAME}?ref=develop" 
        "${CMAKE_BINARY_DIR}/${SUBMODULE_NAME}_info.json"
    )
    file(READ "${CMAKE_BINARY_DIR}/${SUBMODULE_NAME}_info.json" JSON_DATA)
    string(REGEX MATCH "\"sha\":[ \t]*\"([a-fA-F0-9]+)\"" _ "${JSON_DATA}")
    if(CMAKE_MATCH_1)
        set(${OUTPUT_VAR} "${CMAKE_MATCH_1}" PARENT_SCOPE)
        message(STATUS "-> Found ${SUBMODULE_NAME} hash: ${CMAKE_MATCH_1}")
    else()
        message(FATAL_ERROR "Failed to parse hash for ${SUBMODULE_NAME}")
    endif()
endfunction()

# Then they can just call this for any dependency!
# get_ufs_submodule_hash("CMakeModules" CMAKEMODULES_REF)
# get_ufs_submodule_hash("stochastic_physics" STOCHASTIC_PHYSICS_REF)
```

NickSzapiro-NOAA: Or next version to fetch it too:
```
# Define a function that handles everything automatically
function(fetch_ufs_submodule SUBMODULE_NAME)
    message(STATUS "Querying ufs-weather-model for ${SUBMODULE_NAME} info...")
    
    # Download the JSON payload from the parent repo
    file(DOWNLOAD 
        "https://api.github.com/repos/ufs-community/ufs-weather-model/contents/${SUBMODULE_NAME}?ref=develop" 
        "${CMAKE_BINARY_DIR}/${SUBMODULE_NAME}_info.json"
        STATUS api_status
    )
    
    # Read the JSON response
    file(READ "${CMAKE_BINARY_DIR}/${SUBMODULE_NAME}_info.json" JSON_DATA)
    
    # 1. Extract the SHA
    string(REGEX MATCH "\"sha\":[ \t]*\"([a-fA-F0-9]+)\"" _ "${JSON_DATA}")
    set(SUBMOD_SHA "${CMAKE_MATCH_1}")
    
    # 2. Extract the Submodule Git URL
    string(REGEX MATCH "\"submodule_git_url\":[ \t]*\"([^\"]+)\"" _ "${JSON_DATA}")
    set(SUBMOD_URL "${CMAKE_MATCH_1}")
    
    # Verify we got both pieces of information
    if(SUBMOD_SHA AND SUBMOD_URL)
        message(STATUS "-> Sourced ${SUBMODULE_NAME}: ${SUBMOD_URL} @ ${SUBMOD_SHA}")
        
        # Pass the extracted links directly into FetchContent!
        include(FetchContent)
        FetchContent_Declare(
            ${SUBMODULE_NAME}
            GIT_REPOSITORY ${SUBMOD_URL}
            GIT_TAG ${SUBMOD_SHA}
            SOURCE_DIR "${CMAKE_SOURCE_DIR}/${SUBMODULE_NAME}"
        )
        FetchContent_MakeAvailable(${SUBMODULE_NAME})
        message(STATUS "${SUBMODULE_NAME} downloaded successfully")
    else()
        message(FATAL_ERROR "Failed to parse API response for ${SUBMODULE_NAME}.\nResponse: ${JSON_DATA}")
    endif()
endfunction()

# --- Now you can fetch ANY submodule dynamically with one line ---

fetch_ufs_submodule("CMakeModules")
fetch_ufs_submodule("stochastic_physics")
```

gspetro-NOAA:
@DeniseWorthen There are changes to atmos_cubed_sphere in this subPR. Are things just out of sync or should there be an atmos_cubed_sphere sub-PR added as a dependency?

DeniseWorthen:
@gspetro-NOAA I just updated

gspetro-NOAA:
Could we get reviews on this sub-PR in order to process https://github.com/ufs-community/ufs-weather-model/pull/3207 this week?

AlexanderRichert-NOAA (approved):
CI changes look good to me

fv3/atmos_model.F90
```diff
@@ -1838,1068 +1846,639 @@ end subroutine update_atmos_chemistry
...
+  enddo
 
-    rc=0
-!
-  end subroutine assign_importdata
-!
-  subroutine setup_inlinedata(fieldName, datar82d, logunit)
+  if (GFS_control%cpl_imp_dbg) then
+    call get_date(atmtime+atmtimestep,iyear,imonth,iday,ihour,iminute,isecond)
+    write(timestring, "(I4.4,'-',I2.2,'-',I2.2,'T',I2.2,':',I2.2,':',I2.2)") iyear,imonth,iday,ihour,iminute,isecond
```
NickSzapiro-NOAA: Colons in filename may have same problem as in https://github.com/ufs-community/ufs-weather-model/issues/3209 (?)

NickSzapiro-NOAA (approved):
Thanks @DeniseWorthen ! 

Makes you wonder if names and meanings of coupled variables in GFS data structures should match standard_name in fd_ufs.yaml

fv3/atmos_model.F90
DeniseWorthen: I think it is possible to have the same issue w/ the colons. The file name is constructed the same as for the ``DumpFields=true`` option; we could change both formats to make the timestring match the restart timestring format (YYYYMMDD.HHMMSS). 

Since both are optional debugging flags, this change would have no impact on baseline test already done.

ci/CMakeLists.txt
DeniseWorthen: @AlexanderRichert-NOAA Do you have any comment re: the storing of hashes here?

DeniseWorthen:
I've now succeeded getting the FB to write out for the HAFS RT case after turning off the nest. I've tested that the FB feature does work (exercising the "merge field if mergeflg=.true." code for SST and Ocean U,V) for non-nested configurations.

gspetro-NOAA:
RTs completed successfully on https://github.com/ufs-community/ufs-weather-model/pull/3207. This PR can be reviewed and then merged.

dpsarmie (approved):

DusanJovic-NOAA (approved):

PR #1104, opened 2026-05-04
Refactor C3P and introduce CCPP interface with MPAS compatibility; migrate to submodule

lisa-bengtsson:
This PR introduces a major refactoring of the community convective cloud parameterization (C3P), including a new CCPP-facing interface that is compatible with shared MPAS code. As part of this effort, the codebase is migrated from a subdirectory into a standalone repository.

- Kept the existing CCPP-facing interface from ccpp-physics/physics/CONV/C3/cu_c3_driver.F90 mostly intact, with corresponding updates to newly added cu_c3_driver.meta.

- Extracted the MPAS-facing driver from module_cu_c3.F90 and renamed it cu_c3_driver_mpas.F90.

- Added the needed module_cu_c3 dependencies to the CCPP metadata/build path.

- Temporarily call initModConvParGF() to initialize default namelist/configuration variables. Longer term, these should likely be exposed through the standard UFS/CCPP namelist mechanism.

- Import only the plume-control variables needed by the CCPP driver loop; the remaining initialized module variables are used internally by CUP_C3.

- Reworked the CCPP driver to transpose CCPP (i,k) arrays into the shared C3 (k,i) layout before calling CUP_C3.

- Reconstructed MPAS-style thermodynamic states in the CCPP driver, including temp_old, temp_new, temp_new_BL, and temp_new_ADV, using dynamics and PBL tendencies where available.

- Added wiring to output PBL t and q tendencies from GFS_PBL_generic_post.F90 and mynnedmf_wrapper_post.F90

- Converted CCPP specific humidity inputs/tendencies to dry-air mixing ratio for consistency with C3.

- Adjusted land/sea-mask handling, surface flux units, surface temperature, LCL/source-parcel perturbations, and related preprocessing to more closely match the MPAS driver.

- Replaced the legacy shallow/deep CCPP call block with a direct CUP_C3 path patterned after the MPAS driver and the shared C3 call structure.

- Kept the existing CCPP postprocessing section, adding mappings from the shared C3 plume-indexed outputs back into the legacy CCPP arrays.

- Preserved the existing OpenACC directives in the CCPP driver where possible, but CUP_C3 itself will need OpenACC/device treatment before this is GPU-ready.

- Renamed relevant C3 source files from .F to .F90 so they compile consistently as free-form Fortran.


ccpp/data/CCPP_typedefs.meta
```diff
@@ -388,6 +388,13 @@
   dimensions = (horizontal_dimension)
   type = real
   kind = kind_phys
+[dummy]
```
grantfirl: I'm guessing that this was for debugging? It can be removed now (here and in the code), right?

gspetro-NOAA:
Could we get reviews on this PR in order to move forward with https://github.com/ufs-community/ufs-weather-model/pull/3221

gspetro-NOAA:
> Could we get reviews on this PR in order to move forward with [ufs-community/ufs-weather-model#3221](https://github.com/ufs-community/ufs-weather-model/pull/3221)

@DusanJovic-NOAA @grantfirl Could you take a look?

DusanJovic-NOAA (approved):

ccpp/data/CCPP_typedefs.meta
BrianCurtis-NOAA: Is this resolved?

lisa-bengtsson: Sorry, I missed this - will address this now.

gspetro-NOAA:
@lisa-bengtsson The CCPP PR has been merged. Please revert .gitmodules and update the ccpp-physics hash to [f922ca6](https://github.com/ufs-community/ccpp-physics/commit/f922ca6a5545a2d490328fc04d8cbee9a3311237).

ccpp/data/GFS_typedefs.meta
```diff
@@ -8940,7 +8940,7 @@
 [cap_suppress]
   standard_name = radar_derived_convection_suppression
   long_name = radar-derived convection suppression
-  units = unitless
+  units = none
```
dpsarmie, line 8943: Should https://github.com/NOAA-EMC/ufsatm/blob/develop/ccpp/data/GFS_typedefs.meta#L2312 and https://github.com/NOAA-EMC/ufsatm/blob/develop/ccpp/data/GFS_typedefs.meta#L8943 also be set to none for consistency? No need to hold up this PR, it can be corrected in another PR if it's too much trouble.

dpsarmie (approved):

ccpp/data/GFS_typedefs.meta
lisa-bengtsson: Yes, I think that would make sense. However, if we change it here we need to also change it in CCPP where those are used, and since ccpp just got merged we would need to open up a new PR. Therefore, I would suggest doing it in a different PR. It seems related to the lake parameterization?

dpsarmie: Yes, lets not hold this up for those changes. I was just making a note and we can address this in a later PR.

grantfirl: Technically, all dimensionless variables should use the following: https://github.com/ESCOMP/ESMStandardNames/blob/main/StandardNamesRules.rst

It should probably be set to 1 rather than `none`. However, it shouldn't be done in this PR for the reasons @lisa-bengtsson articulated. Plus, it looks like that particular variable has been orphaned by the CLM lake model anyway, so the entire variable and its metadata should be removed.

gspetro-NOAA:
This PR can be merged. Sub-PRs have been merged, .gitmodules reverted, and the physics hash updated.

ccpp/data/GFS_typedefs.meta
lisa-bengtsson: @grantfirl yes - maybe "unit consistency" is something to add to the house keeping PR we've been discussing.

PR #1135, opened 2026-09-01
Add new compiler options to ThompsonMP for speed-up

### LINKED ISSUE #3347: Add Thompson MP compile options for performance speed-up
## Description
HAFS has tested compiler options for a speed up in the Thompson MP scheme (https://github.com/hafs-community/ccpp-physics/blob/production/hafs.v2.2/CMakeLists.txt#L87) that will be a part of the HAFSv2.2 production suite.  The GFS team also tested these changes and did not see anything that would prevent them from using these options in the future. 
These options will NOT be in GFSv17 because these options were made available too late in the model process, but it could be added in future iterations.

These options should be added to develop. 

## Solution
Add compile options into the new ccpp capgen framework.

dpsarmie:
## Description

This PR will add new compiler options that were found to speed up the Thompson microphysics scheme by ~3%-4%. These options will change baselines, but they are within the expected range according to work done by the HAFS and GFS teams.

These compiler options are going to be in HAFSv2.2 production.

### Issue(s) addressed

- fixes https://github.com/ufs-community/ufs-weather-model/issues/3347

## Testing

These changes were tested with the UFS WM regression test system.  Baseline changes were documented and reproducibility was tested on Ursa.

## Dependencies

None

### DIFF
--- a/ccpp/CMakeLists.txt
+++ b/ccpp/CMakeLists.txt
@@ -467,6 +467,17 @@ if (CMAKE_BUILD_TYPE STREQUAL "Release" AND (${CMAKE_Fortran_COMPILER_ID} STREQU
 endif()
 
 #------------------------------------------------------------------------------
+# Thompson MP can be sped up with more aggresive compiler flags;
+# HAFS will have this in 2.2 production. This is Intel/IntelLLVM only.
+#------------------------------------------------------------------------------
+if (CMAKE_BUILD_TYPE STREQUAL "Release" AND (${CMAKE_Fortran_COMPILER_ID} STREQUAL "Intel" OR ${CMAKE_Fortran_COMPILER_ID} STREQUAL "IntelLLVM"))
+	message(STATUS "Add speed up flags to Thompson MP with Intel")
+  set_property(SOURCE
+    ${SCHEME_FILES_DIR}/physics/MP/Thompson/module_mp_thompson.F90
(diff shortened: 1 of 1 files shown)


ccpp/CMakeLists.txt
```diff
@@ -466,6 +466,17 @@ if (CMAKE_BUILD_TYPE STREQUAL "Release" AND (${CMAKE_Fortran_COMPILER_ID} STREQU
     APPEND PROPERTY COMPILE_OPTIONS -O1)
 endif()
 
+#------------------------------------------------------------------------------
+# Thompson MP can be sped up with more aggresive compiler flags;
+# HAFS will have this in 2.2 production. This is Intel/IntelLLVM only.
+#------------------------------------------------------------------------------
+if (CMAKE_BUILD_TYPE STREQUAL "Release" AND (${CMAKE_Fortran_COMPILER_ID} STREQUAL "Intel" OR ${CMAKE_Fortran_COMPILER_ID} STREQUAL "IntelLLVM"))
```
NickSzapiro-NOAA, line 473: Previous discussions like https://github.com/ufs-community/ccpp-physics/issues/86 and https://github.com/ufs-community/ccpp-physics/pull/76 seem to favor putting this behind a "FASTER" option or such ... so 1) applications must explicitly opt into this behavior they confirm works for them and 2) default release maintains reproducibility

